PR #11, opened 2024-01-11
Move Decapodes ACSet to InterTypes file

olynch:
Supersedes https://github.com/AlgebraicJulia/Decapodes.jl/pull/184

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
 CombinatorialSpaces = "b1c52339-7909-45ad-8b6a-6e388f7c67f2"
+DiagrammaticEquations = "6f00c28b-6bed-4403-80fa-30e0dc12f317"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MLStyle = "d8e11817-5142-5d16-987a-aa16d5891078"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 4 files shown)


test/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
 CombinatorialSpaces = "b1c52339-7909-45ad-8b6a-6e388f7c67f2"
+DiagrammaticEquations = "6f00c28b-6bed-4403-80fa-30e0dc12f317"
```
jpfairbanks, line 4: Are you supposed to include a package in its own test deps? How were the test passing before?

jpfairbanks (approved):
Looks good to me. Maybe an unnecessary addition of package to its own test deps. But we can keep this branch moving forward while Decapodes adopts DE.jl and then update Decapodes.jl to use inter types too.

jpfairbanks:
@quffaro I can't see why the docs are failing on buildkite. The build log is empty when I log in on my ipad both safari and chrome. A possible reason would be that the docs Project.tool needs to be updated to use ACSets.jl Or maybe some display methods are missing on InterTypes objects.

test/Project.toml
olynch: I honestly don't know. Normally I run `julia --project=test` and then `]develop .` before running my tests, and this is what got added when I did that.

olynch:
@jpfairbanks should we merge this despite the docs failing? What's the action item here.

quffaro:
@olynch I've determined the root cause of the issue.

When you pushed your commit, Buildkite was running two agents: one on the daemon node and one on a local Macbook (mine?). The issue was that the agent running on my Macbook picked up the job to run in its own buildkite directory, so it failed when the "module" function was not present.

I stopped that agent so the daemon was remained and retried the build. It appears to be working now.

PR #14, opened 2024-01-19
CompatHelper: add new compat entry for ACSets at version 0.2, (keep existing compat)

algebraicjuliabot:
This pull request sets the compat entry for the `ACSets` package to `0.2`.
This keeps the compat entries for earlier versions.

Note: I have not tested your package with this new compat entry.
It is your responsibility to make sure that your package tests pass before you merge this pull request.
Note: Consider registering a new release of your package immediately after merging this PR, as downstream packages may depend on this for tests to pass.

PR #22, opened 2024-02-02
Add ACSets@0.3 to compat

LilithHafner:
Originally posted by @lukem12345 in https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/14/commits/d25683c8fcf09f274690196b33d9dadff8b0195a.

I don't know anything about whether DiagrammaticEquations is actually compatible with ACSets 0.3, but I want to make sure that change doesn't slip through the cracks.


lukem12345:
Hey, ACSets 0.3 isn’t out yet. That change isn’t coordinated with AlgebraicRewriting

PR #26, opened 2024-04-30
Check black list when contracting operators

lukem12345:
The `contract_operators` function finds chains of unary operators, and replaces them with a single operator containing a vector of the operators from that chain.

We currently support the ability to only contract when operators are in some given white list of operators. When there is no white list given, we default to allowing any operator to be chained.

It will simplify future Decapode manipulation if we also support a black list of operators. If an operator is in the black list, such an operator will not be chained.

### DIFF
--- a/src/acset.jl
+++ b/src/acset.jl
@@ -299,7 +299,9 @@ function contract_operators(d::SummationDecapode;
   #return e
 end
 
-function contract_operators!(d::SummationDecapode; white_list::Set{Symbol} = Set{Symbol}(), black_list::Set{Symbol} = Set{Symbol}())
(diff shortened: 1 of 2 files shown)


lukem12345:
### Before
```mermaid
flowchart LR

A -->|g| B
B -->|f| C
C -->|e| D
D -->|d| E
E -->|c| F
F -->|b| G
G -->|a| H

```
```julia
contract_operators(t15_orig,
    white_list=Set([:a, :b, :c]),
    black_list=Set([:d]))
```
### After
```mermaid
flowchart LR

A -->|g| B
B -->|f| C
C -->|e| D
D -->|d| E
E -->|c,b,a| H

```

PR #37, opened 2024-06-07
Use Match-Replace strategy to alias operators

### LINKED ISSUE #35: Better support for using Operator Aliases
The goal here is to use familiar `Div, Grad, Curl, Lapl` operators in their familiar role in vector calculus. What support do we need in the Decapodes language for that? Probably and alias operator like `:=`

```julia
oscillator = @decapode begin
  X::Form0
  V::Form0
  div := ⋆⋅d
  
  ∂ₜ(X) == V
  ∂ₜ(V) == -k(X) + div(V)
end
```

Then we can update all the docs examples to use these operators when appropriate. We can't uniformly introduce a exterior calculus definitions of div, grad, curl because of things like how gradient of Pressure should be `dP` but gradient of concentration should be `⋆dC`. Unless we can dispatch on expected return type in the type inference code, I think that we can just introduce a manual aliasing operator into the decapode macro, and then have local aliases in each example.

lukem12345:
Close #35 

As a compromise between:
- considerations of computational efficiency
- warts relating to the types of "matched" variables destroying "context"
, a "match & replace" strategy is adopted - inspired by AlgebraicRewriting - but using tight loops tailored to our workflow.


lukem12345:
I'll stick a pin in [03aa7ce](https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/37/commits/03aa7ce18fce54e9f81f3cb55b1ec4708b3e9d11) as a complete version of this feature in an imperative/ database-manipulation manner.

I'll continue commits to this branch from the operadic composition approach.

jpfairbanks:
Let's keep the other approach in separate files so that we can compare and contrast the changes. I'm thinking the repurposing of oapply will be much more maintainable long term.

lukem12345:
Record of discussion with @GeorgeR227 :
- Take care that the RHS Decapodes are fully expanded
- Reformat (use newlines) in replace-operator functions
- Merge the "more imperative" code (after review process) in 03aa7ce. (It is more declarative than existing code currently in Decapodes.)
- Decide whether we should export the single-operator replacement functions.
- Even the oapply approach will need some imperativity on top of it, either creating RHS Decapodes on-the-fly with the names of the src/tgt variables, or in the LHS, and the replaced operator still has to be explicitly removed.

lukem12345:
Default composition code has been cherry picked onto PR #52 . Upon that merging, I will reset this PR to 03aa7c

Issue #43, opened 2024-06-17
Reexport @relation


GeorgeR227:
Even though DiagrammaticEquations exports its own composition code, like `oapply`, it does not currently reexport the `@relation` macro that downstream packages would need to actually compose. As of now, a user would have to add Catlab separately to have this work, which is not at all clear. However, DiagrammaticEquations already requires Catlab to function so reexporting `@relation` can be done at no cost.

lukem12345:
Would updating documentation also close this?

GeorgeR227:
I would say that since `@relation`, plus `apex` as I've found, is required to use the composition code here, the best solution is to just reexport it. I already have a branch set up.

PR #44, opened 2024-06-17
Reexport Catlab functionality

GeorgeR227:
Added `apex` from DiagrammaticEquations and `@relation`, `to_graphviz` from Catlab to exports. These are required to use the composition code and the visualization code.


GeorgeR227:
Close #43

lukem12345:
This might be tailoring a little too close to the happy path. What about when we want to compose Decapodes with `oplus`. Why only export some composition functionality from Catlab but not all?

@jpfairbanks thoughts?

GeorgeR227:
I was just adding what I thought were the most often used functions. We can add some more if we want but if we start re-exporting too much then at that point it'd be better to just get the user to have Catlab.

jpfairbanks (approved):

Issue #46, opened 2024-06-20
Bad typing for Decapodes


GeorgeR227:
This bug seems to have been introduced in version 0.1.5 and is causing the Decapodes documentation to break. Below is the error caused.
```julia
Error: failed to run `@example` block in src/ice_dynamics/ice_dynamics.md:200-203
│ ```@example DEC
│ sim = eval(gensim(ice_dynamics1D, dimension=1))
│ fₘ = sim(sd, generate)
│ ```
│   exception =
│    ArgumentError: cannot parse "dynamics_•2" as Float64
```

I've confirmed this is due to 0.1.5 since I've bumped the version back down to 0.1.4 and the same page built fine. This will be resolved once the issue is found and tests are added to prevent this issue again.

lukem12345:
I suspect it is because types are inferred in order of precedence as given in the type inference rules list. This has come up before. Adding a new operator to the list changes that order.

lukem12345:
So, a rule which infers literals would be a likely cause.

lukem12345:
So, a rule which infers literals would be a likely cause.

Due to the 2 out of 3 logic for matching, lines like this one can cause trouble:

```julia
  (proj1_type = :Literal, proj2_type = :Form1, res_type = :Form1, op_names = [:/, :./, :*, :.*, :^, :.^]),
```

lukem12345:
So, a rule which infers literals would be a likely cause.

Due to the 2 out of 3 logic for matching, lines like this one can cause trouble:

```julia
  (proj1_type = :Literal, proj2_type = :Form1, res_type = :Form1, op_names = [:/, :./, :*, :.*, :^, :.^]),
```

This line that just checks whether 2 out of 3 are `:infer` should be amended. Literals have to have special treatment.
https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/blob/main/src%2Facset.jl#L412

GeorgeR227:
Yeah, looking at the Decapode itself, the `dynamics_•2` in question is the result of an `avg_01`, which is then the source of a `*`. I'm seeing now this issue didn't actually appear in this version was was uncovered instead.

<img width="384" alt="image" src="https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/assets/78235421/bf3fea9c-bb16-4469-9733-013108bf82fb">

This voting logic seems to be coming back to bite us again and again.

lukem12345:
[...]

PR #47, opened 2024-06-20
Safer type modification

GeorgeR227:
This should help solve problems in type inference regarding typing rules doing strange things. Addresses problems but does not fully resolve issue #46.

### DIFF
--- a/src/acset.jl
+++ b/src/acset.jl
@@ -341,13 +341,25 @@ function add_parameter(d::AbstractNamedDecapode, k::Symbol)
     return add_part!(d, :Var, type=:Parameter, name=k)
 end
 
-function safe_modifytype(tgt_type::Symbol, src_type::Symbol)
-  modify = (tgt_type == :infer && !(src_type == :Literal || src_type == :Constant || src_type == :Parameter))
-  return (modify, modify ? src_type : tgt_type)
+
+"""
+    safe_modifytype(org_type::Symbol, new_type::Symbol)
+
+This function accepts an original type and a new type and determines if the original type
+  can be safely overwritten by the new type.
+"""
+function safe_modifytype(org_type::Symbol, new_type::Symbol)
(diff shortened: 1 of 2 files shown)


GeorgeR227:
To be clear, this should resolve the issues present in that issue but a MWE and proper tests are still not available.

lukem12345:
I would change it such that `safe_modifytype` is mutating, and just returns true or false, unless its too unseemly to pass the necessary ref. 

Adding the Halfar equation with `avg_01` ought to suffice for now, and the code cov report is an automatic check to confirm that this is getting hit.

GeorgeR227:
I decided to make a modifying wrapper around the function. Reason being that I want the core of the function itself to be very short and concise, plus leaving it like this makes testing it easier since I don't need to build up Decapodes for it.

src/acset.jl
```diff
@@ -341,7 +341,18 @@ function add_parameter(d::AbstractNamedDecapode, k::Symbol)
     return add_part!(d, :Var, type=:Parameter, name=k)
 end
 
+function safe_modifytype(tgt_type::Symbol, src_type::Symbol)
```
lukem12345, line 344: A better name should be picked instead of using "src" and "tgt". Adding a comment above this line will help clarify as well.

GeorgeR227: I've changed this to be `org_type` and `new_type` respectively, plus added docstrings. I've looked at too much assembly before I guess.

lukem12345: The problem is src and tgt are just too good and then get used everywhere

lukem12345:
Want to merge this now and do a separate PR for Constant?

GeorgeR227:
Yeah that's the plan. The Constant behavior is baked into the type inference so changing that seems more involved.

lukem12345 (approved):

GeorgeR227:
Close #46. We now have a MWE and a test to cover it.

PR #48, opened 2024-06-21
Work on summation inference

GeorgeR227:
Cleans up the summation inference code. This will now only propagate form types and do basic type checking.

### DIFF
--- a/src/acset.jl
+++ b/src/acset.jl
@@ -354,45 +354,42 @@ function safe_modifytype(org_type::Symbol, new_type::Symbol)
 end
 
 """
-    safe_modifytype!(d::SummationDecapode, var_idx::Int, org_type::Symbol, new_type::Symbol)
+    safe_modifytype!(d::SummationDecapode, var_idx::Int, new_type::Symbol)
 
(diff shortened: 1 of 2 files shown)


lukem12345:
@GeorgeR227 Your last few PRs have had spurious changes to lines. I guess this is because of line ending encodings or something. I think you can configure your git client to ignore these.

GeorgeR227:
I think actually this is to do with hanging whitespace at the end of lines. My git client does tell me if the encoding are different. 

However I agree with you in that these are annoying now because they aren't the purpose of these PRs.

GeorgeR227:
Also I'm moving this PR to #49.

lukem12345:
> I think actually this is to do with hanging whitespace at the end of lines. My git client does tell me if the encoding are different. 

Yeah maybe spot check a few of these. It usually highlights when whitespace inside of line is added or removed.

It makes the git blame misleading when tracking things down.

lukem12345:
Ok, maybe turn off the aggressive linting setting or just use `git add` at the command line instead.

GeorgeR227:
I'll open a linting issue with a reference this problem. We have an issue about this in Decapodes but not here.

PR #49, opened 2024-06-21
 Work on summation inference

GeorgeR227:
Cleans up the summation inference code. This will now only propagate form types and do basic type checking.

### DIFF
--- a/src/acset.jl
+++ b/src/acset.jl
@@ -454,7 +454,7 @@ function infer_types!(d::SummationDecapode, op1_rules::Vector{NamedTuple{(:src_t
   types_known_op1[incident(d, :infer, [:src, :type])] .= false
   types_known_op1[incident(d, :infer, [:tgt, :type])] .= false
 
-  types_known_op2 = zeros(Bool, nparts(d, :Op2))
+  types_known_op2 = ones(Bool, nparts(d, :Op2))
   types_known_op2[incident(d, :infer, [:proj1, :type])] .= false
   types_known_op2[incident(d, :infer, [:proj2, :type])] .= false
   types_known_op2[incident(d, :infer, [:res, :type])] .= false
--- a/src/acset.jl
+++ b/src/acset.jl
@@ -373,28 +373,29 @@ function filterfor_forms(types::AbstractVector{Symbol})
   filter(conditions, types)
 end
 
-function infer_summands_and_summations!(d::SummationDecapode)
+function infer_summands_and_summations!(d::SummationDecapode, Σ_idx::Int)
   # Note that we are not doing any type checking here for users!
   # i.e. We are not checking the underlying types of Constant or Parameter
   applied = false
 
-  for Σ_idx in parts(d, :Σ)
-    summands = d[incident(d, Σ_idx, :summation), :summand]
-    sum = d[Σ_idx, :sum]
-    idxs = [summands; sum]
+  summands = d[incident(d, Σ_idx, :summation), :summand]
+  sum = d[Σ_idx, :sum]
+  idxs = [summands; sum]
+  types = d[idxs, :type]
+  all(t != :infer for t in types) && return applied # We need not infer
 
-    forms = unique(filterfor_forms(d[idxs, :type]))
+  forms = unique(filterfor_forms(types))
 
-    form = @match length(forms) begin
-      0 => continue # We need not infer, We can not infer
-      1 => only(forms)
-      _ => error("Type mismatch in summation")
-    end
-
-    for idx in idxs
-      applied |= safe_modifytype!(d, idx, form)
-    end
+  form = @match length(forms) begin
+    0 => return applied # We can not infer
+    1 => only(forms)
+    _ => error("Type mismatch in summation")
   end
+
+  for idx in idxs
+    applied |= safe_modifytype!(d, idx, form)
+  end
+
   return applied
 end
 
@@ -468,7 +469,7 @@ function infer_types!(d::SummationDecapode, op1_rules::Vector{NamedTuple{(:src_t
         this_applied = apply_inference_rule_op1!(d, op1_idx, rule)
 
         types_known_op1[op1_idx] = this_applied
-        applied = applied || this_applied
+        applied |= this_applied
       end
     end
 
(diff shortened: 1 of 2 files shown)


src/acset.jl
```diff
@@ -354,46 +354,48 @@ function safe_modifytype(org_type::Symbol, new_type::Symbol)
...
-function infer_summands_and_summations!(d::SummationDecapode)
-  # Note that we are not doing any type checking here!
-  # i.e. We are not checking for this: [Form0, Form1, Form0].
+"""
+    filterfor_forms(types::AbstractVector{Symbol})
+
+Return any form type symbols.
+"""
+function filterfor_forms(types::AbstractVector{Symbol})
+  conditions = x -> x != :Literal && x != :Constant && x != :Parameter && x != :infer
```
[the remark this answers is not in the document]
GeorgeR227: Changed but I think eventually changing these symbols to Julia types is the way to go.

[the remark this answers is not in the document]
lukem12345: Yeah agreed

test/language.jl
```diff
@@ -356,6 +356,45 @@ end
...
 
+import DiagrammaticEquations: @get_alltypes, @get_formtypes, @get_primalformtypes, @get_dualformtypes,
+@get_nonformtypes, @get_usertypes, @get_numbertypes, @get_noninferabletypes, @get_infertypes
+@testset "Type Retrival" begin
+  calls = [:@get_alltypes, :@get_formtypes, :@get_primalformtypes, :@get_dualformtypes,
+  :@get_nonformtypes, :@get_usertypes, :@get_numbertypes, :@get_noninferabletypes, :@get_infertypes]
+
+  # No repeated types
+  for call in calls
+    res = @eval $(call)
```
lukem12345, line 367: I get that it is some sort of compromise for not having the types, and it is interesting, but this metaprogramming needs to go.

test/language.jl
```diff
@@ -356,6 +356,45 @@ end
...
+    @test unique(res) == res
```
lukem12345, line 368: There is an `allunique` function that can be used here.

src/acset.jl
```diff
@@ -155,13 +155,33 @@ function make_sum_mult_unique!(d::AbstractNamedDecapode)
   end
 end
 
+# A collection of DecaType getters
+# TODO: This should be replaced by using a type hierarchy
+macro get_alltypes()
```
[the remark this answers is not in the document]
lukem12345: Just use constant vectors or sets for the time being. This is certainly an interesting way of not using types directly, but this is quite strange.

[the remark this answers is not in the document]
GeorgeR227: Yeah you're right, constant vectors would be the way to go.

test/language.jl
GeorgeR227: Yeah idk why I got stuck on the macro way, constants are better here. But this is meant to all test how we should organize our types hierarchically when do make that switch.

test/language.jl
```diff
@@ -742,6 +779,56 @@ end
...
+      D::Constant
+      A == C + D + B
+    end
+    infer_types!(d)
+
+    names_types_expected = Set([(:A, :Form0), (:B, :Form0), (:C, :Parameter), (:D, :Constant)])
+    @test test_nametype_equality(d, names_types_expected)
+  end
+
+  # Test #24: Check that different forms summed up error out
```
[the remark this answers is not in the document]
GeorgeR227: I've changed the comment to this.

src/acset.jl
```diff
@@ -349,51 +368,53 @@ This function accepts an original type and a new type and determines if the orig
...
-function infer_summands_and_summations!(d::SummationDecapode)
-  # Note that we are not doing any type checking here!
-  # i.e. We are not checking for this: [Form0, Form1, Form0].
+"""
+    filterfor_forms(types::AbstractVector{Symbol})
+
+Return any form type symbols.
+"""
+function filterfor_forms(types::AbstractVector{Symbol})
+  conditions = x -> !(x in NONFORM_TYPES)
```
lukem12345, line 391: Feels more natural to use `in FORM_TYPES` here.

src/acset.jl
```diff
@@ -349,51 +368,53 @@ This function accepts an original type and a new type and determines if the orig
...
+
+  forms = unique(filterfor_forms(types))
+
+  form = @match length(forms) begin
+    0 => return applied # We can not infer
+    1 => only(forms)
+    _ => error("Type mismatch in summation $Σ_idx, all the following forms appear: $forms")
   end
+
+  for idx in idxs
```
lukem12345, line 414:
```julia
applied = any(map(x -> safe_modifytype!(d, x, form), idxs))

File: src/acset.jl [hunk 8]
```diff
@@ -349,51 +368,53 @@ This function accepts an original type and a new type and determines if the orig
...
+"""
+function filterfor_forms(types::AbstractVector{Symbol})
+  conditions = x -> !(x in NONFORM_TYPES)
+  filter(conditions, types)
+end
+
+function infer_sum_types!(d::SummationDecapode, Σ_idx::Int)
+  # Note that we are not doing any type checking here for users!
+  # i.e. We are not checking the underlying types of Constant or Parameter
   applied = false
```
lukem12345, line 398: Remove this line because subsumed by next review comment.

src/acset.jl
```diff
@@ -155,13 +155,32 @@ function make_sum_mult_unique!(d::AbstractNamedDecapode)
...
+
+const NONFORM_TYPES = [:Constant, :Parameter, :Literal, :infer]
+const USER_TYPES = [:Constant, :Parameter]
+const NUMBER_TYPES = [:Literal]
+const INFER_TYPES = [:infer]
+
+# Types that can not ever be inferred
+const NONINFERABLE_TYPES = [:Constant, :Parameter, :Literal]
+
+function get_unsupportedtypes(types)
```
lukem12345, line 175: Can we set the type sig here to be an abstract vector of symbols?

GeorgeR227: Unfortunately we can't since Decapodes think our types are `Any`, not `Symbol`.

src/acset.jl
GeorgeR227: Agreed

src/acset.jl
GeorgeR227: Nice

GeorgeR227: I'll leave the `applied` var name just for programmer clarity.

src/acset.jl
GeorgeR227: I forget why we had to settle for `Decapode{Any, Any, Symbol}` but maybe we can tackle this in a new PR.

lukem12345: Yeah that's annoying.

We should re-consider
https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/blob/main/src%2Flanguage.jl#L197
To standardize on symbol symbol symbol

lukem12345:
We can clean up the setting of the `applied` variable by using helper functions for Op1 and Op2 inference like we do for Summations, but it's ok for now.

lukem12345 (approved):

PR #52, opened 2024-06-24
Default Composition Diagrams

### LINKED ISSUE #36: Default Composition
Currently, composing Decapodes requires a user to both construct a UWD and specify internal variables when oapplying.

However, some users already have a composition in-mind  when constructing components, and already "globally namespace" their variables.

So, we should offer a helper method that constructs a default UWD that assumes that variables with the same `name` are to be composed, and oapplies this UWD.

lukem12345:
Close #36

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -3,24 +3,8 @@
 import Catlab.CategoricalAlgebra: apex, feet, legs
 import Catlab.WiringDiagrams: oapply
 
-OpenSummationDecapodeOb, OpenSummationDecapode = OpenACSetTypes(SummationDecapode, :Var)
-
-#FIXME: why can't we just add a constructor for OpenSummationDecapode
-"""    Open(d::SummationDecapode{T,U,V}, names::AbstractVector{Symbol}) where {T,U,V}
-
-creates an OpenSummationDecapode based on named variables rather than variable indices. 
-See AlgebraicPetri.jl's Open for the analogous verion for LabelledReactionNetworks.
-"""
-function Open(d::SummationDecapode{T,U,V}, names::AbstractVector{Symbol}) where {T,U,V}
-  legs = map(names) do name
-    FinFunction(incident(d, name, :name), nparts(d, :Var))
-  end
-  OpenSummationDecapode{T,U,V}(d, legs...)
-end
-
-apex(Decapode::OpenSummationDecapode) = apex(Decapode.cospan)
-legs(Decapode::OpenSummationDecapode) = legs(Decapode.cospan)
-feet(Decapode::OpenSummationDecapode) = Decapode.feet
+# ACSet manipulation helper functions
+# -----------------------------------
 
 """    function unique_by!(acset, column_names::Vector{Symbol})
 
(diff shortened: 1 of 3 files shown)


src/composition.jl
```diff
@@ -212,3 +216,44 @@ oapply(r::RelationDiagram, podes::Vector{D}) where {D<:OpenSummationDecapode} =
...
+
+Note that composing immediately with [`oapply`](@ref) will fail if types do not match (e.g. (:infer, :Form0) or (:Form0, :Form1)).
+"""
+function default_composition_diagram(podes::Vector{D}, names::Vector{Symbol}, only_states_terminals=false) where {D<:SummationDecapode}
+  non_lit_names = map(podes) do pode
+    pode[findall(!=(:Literal), pode[:type]), :name]
+  end
+  for (nln, name) in zip(non_lit_names, names)
+    allunique(nln) || error("Decapode $name contains repeated variable names: $nln.")
+  end
```
GeorgeR227, line 250: Change the error to output just the specific names that are repeated. For large Decapodes it'll be too much information to sift through.

src/composition.jl
```diff
@@ -212,3 +216,44 @@ oapply(r::RelationDiagram, podes::Vector{D}) where {D<:OpenSummationDecapode} =
 # oapply(r, OpenPode(Heat, [:H]))
 
 oapply(r::RelationDiagram, pode::OpenSummationDecapode) = oapply(r, [pode])
+
+# Default composition
+# -------------------
+
+# TODO: Upstream this to Catlab?
+function construct_relation_diagram(boxes::Vector{Symbol}, junctions::Vector{Vector{Symbol}})
```
GeorgeR227, line 224: Have a pair of a box and its junctions be a struct, then have this function accept an `AbstractVector{struct}`. Checking for type `Vector{Vector{Symbol}}` is never a good sign.

src/composition.jl
```diff
@@ -212,3 +216,44 @@ oapply(r::RelationDiagram, podes::Vector{D}) where {D<:OpenSummationDecapode} =
...
+  tables = map(boxes, junctions) do b, j
+    Expr(:call, b, j...)
```
GeorgeR227, line 226: If you use a struct then you can overload `Base.expr` for it, like we do in Decapodes for `AbstractCalls`.

test/composition.jl
```diff
@@ -195,4 +195,60 @@ end
...
+  end
+  expected =
+    @relation () begin
+      Diffusion(C,ϕ₁)
+      Advection(C,V,ϕ₂)
+      Superposition(C,ϕ₁,ϕ₂,Ċ,ϕ)
+    end
+  @test is_isomorphic(expected, default_composition_diagram(
+          [Diffusion, Advection, Superposition],
+          [:Diffusion, :Advection, :Superposition]))
```
GeorgeR227, line 227: Feels like we should be able to just use the name of the Decapodes themselves but not sure how to implement that.

lukem12345: Yeah no worries there is a TODO for that # TODO: Add a macro which provides names for boxes via the Symbol of the Decapode.

src/composition.jl
lukem12345: I implemented this locally, but it just ended up adding a lot of logic to wrap and unwrap values into and out of the struct, increasing noise. I think if there were other places where a box-junctions pair needs to be represented, or more was done with the pair here than just being passed to a single call to Expr, then we can take a look at this again.

src/composition.jl
```diff
@@ -212,3 +216,51 @@ oapply(r::RelationDiagram, podes::Vector{D}) where {D<:OpenSummationDecapode} =
...
+
+No Literals are exposed. Use [`unique_lits!`](@ref) after composing.
+
+Throw an error if any individual Decapode already contains a repeated name (except for Literals).
+
+If `only_states_terminals` is `true`, only expose state and terminal variables. Defaults to `false`.
+
+Note that composing immediately with [`oapply`](@ref) will fail if types do not match (e.g. (:infer, :Form0) or (:Form0, :Form1)).
+"""
+function default_composition_diagram(podes::Vector{D}, names::Vector{Symbol}, only_states_terminals=false) where {D<:SummationDecapode}
```
GeorgeR227, line 251: Add a check here to confirm the length of `podes` and `names` is the same and throw an error otherwise.

test/composition.jl
```diff
@@ -195,4 +195,60 @@ end
 # Test none of the decapodes were mutated
 @test isequal(adv_adv, deep_copies)
 
+@testset "Default Composition Diagrams" begin
```
GeorgeR227, line 198: Add a test to check that the errors are correctly thrown. So one for length check and one for dup name check.

Issue #34, opened 2024-06-06
Convert all Unicode to Ascii


GeorgeR227:
In Decapodes specifically but perhaps more generally as well, we should make all code relying on symbol name matching to depend on matching to the Ascii version of that symbol name. Unicode names can be supported but under the hood these should be converted to Ascii when needed. 

Doing this makes the underlying code more maintainable since we only ever need to worry about the Ascii name and having the Ascii name be the base makes the code more accessible to users.

jpfairbanks:
Having better ways to centralize the Unicode and ascii aliases would be good. But we can't just always convert to one or the other. If the users wrote it in Unicode then they will want to see output with Unicode labels on plots and stuff.

GeorgeR227:
Yeah I understand, which is why this conversion won't be happening directly on user facing things, like a user-made ACSet, but either on back-end copies of that ACSet or with on-the-spot conversions. 

So for example, when we do type-inference, what we do now is look for multiple variations of an operator name.
https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/blob/fa4fe1e9f2d8afac3086c8453a14550507c82305/src/deca/deca_acset.jl#L9-L14

The idea would be to instead convert these names within the type-inference function itself, so hidden to the user, to a base name, represented in Ascii, and then have these rules only match on that base name.

lukem12345:
Yeah we can come up with some useful wrappers around symbols, tagging them as "USER_INPUTTED" and that kind of thing. For recognized symbols (or some distinct class of symbol wrappers) we can emit different code at `to_graphviz` time vs. colanguage time vs. `gensim` time.

lukem12345:
Sufficiently advanced feature would be able to track the lifetime as a symbol behind the scenes. Storing information such as:

"This variable was originally from the Foo decapode with name Bar. It was composed with Biz from Baz and called Buzz."

Or

"This variable was created by calling `expand_operators` on Baz and called Biz, then composed with Bar from the Foo decapode."

i.e. Some of the tools needed to provide a "stacktrace" or debugging information

PR #54, opened 2024-06-25
Add operator name canonicalization

GeorgeR227:
This is meant to address #34, to allow for better support of using Unicode and Ascii operator names. The idea here is to allow the user to choose from a range of supported operator names but have the underlying codebase only work on the canon name.

An example would be to have type inference rules carry a canon name, instead of an array of supported names, and have inference rules simply check that their name matches the converted user name.

These canon names should be carefully chosen to be easy to work with and parse. Some rules are included but are liable to change.


test/op_naming.jl
```diff
@@ -0,0 +1,600 @@
...
+      R3 == i_1(D, B)
+      R4 == i₂(D, C)
+    end
+    check_canontyping(gen_d, d)
+    check_canonoverload_op2(gen_d, d)
+  end
+end
+
+@testset "Typed Function Observance" begin
+  let # Typing respects typed exterior derivative
```
lukem12345, line 371: What is this test supposed to show?

test/op_naming.jl
```diff
@@ -0,0 +1,600 @@
+using Test
+using DiagrammaticEquations
+
+@testset "Canon Inference and Overloading" begin
+  function check_canontyping(control::SummationDecapode, test::SummationDecapode)
+    infer_test = infer_types!(deepcopy(test))
```
lukem12345, line 6: Can we use `infer_types!(deepcopy(test)) == infer_types!(deepcopy(control))`

test/op_naming.jl
GeorgeR227: When I first implemented canon name typing, if a user typed in `A == d_0(B)` but said `B::Form1`, which is wrong cause the operator doesn't match with the input, infer would see `d_0` as `d`, the canon generic name, and type `A` as a `Form2`. 

When I realized that, I changed the code so if the operator is already typed it can only use its own specific rule and not the generic rules. These tests enforce this behavior because it felt like what was occurring above is a silent failing, leading to bad errors down the line.

test/op_naming.jl
```diff
@@ -0,0 +1,600 @@
...
+    @test infer_test[:op2] == test[:op2]
+  end
+
+  function check_canonoverload_op1(control::SummationDecapode, test::SummationDecapode)
+    over_test = resolve_overloads!(infer_types!(deepcopy(test)))
+    over_control = resolve_overloads!(infer_types!(deepcopy(control)))
+    @test get_canon_name(over_test[:op1]) == get_canon_name(over_control[:op1])
+  end
+
+  setup_basecase(d::SummationDecapode) = resolve_overloads!(infer_types!(deepcopy(d)))
```
lukem12345, line 19: "Setup" is too generic of a function name here, and it looks like deepcopy, infer, resolve is used in the function above this.

test/op_naming.jl
GeorgeR227: Technically the scope of this test is just to checks types and that names don't change. I guess we could change it to that but I want to limit the scope of the test to test what we actually want to check.

src/deca/deca_op_names.jl
lukem12345, line 1: This file should get renamed to `theory.jl`. Later, we can silo all of the information that would normally incorporate `@theory` or "Model" information. The reason we won't use `@theory` outright for the time being is that we want to have support for multiple names referring to the same operator, etc.

test/op_naming.jl
lukem12345: Ok. We should have a comment that explains this situation using e.g. that above summary. The "respects" feels a little overloaded.

lukem12345: Do we need to test this behavior for all the other operators as well, or does this single test suffice?

lukem12345:
This PR just uses the global names Dict for the type inference and function overloading, but it should be used throughout this codebase and Decapodes.jl wherever explicit symbols appear. Can you spot check the rest of this codebase for such updates? Where it is too tedious (e.g. `unicode!`), we can make an issue to refactor.

src/deca/deca_op_names.jl
```diff
@@ -0,0 +1,254 @@
...
+# Op 1 names
+const PARTIAL_T = :dt
+
+const EXTDERIV_0 = :d_0
+const EXTDERIV_1 = :d_1
+
+const DUALDERIV_0 = :dual_d_0
+const DUALDERIV_1 = :dual_d_1
+
+const HODGE_0 = :hdg_0
```
lukem12345, line 28: The ASCII-Unicode equivalents that we have now use `star` and `star_inv` instead of `hdg`.

https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/blob/e21e6c7b61a2158a912247af5b0121785046a02c/src/deca/deca_acset.jl#L281

src/deca/deca_op_names.jl
```diff
@@ -0,0 +1,254 @@
...
+# ⋆₀, hdg_0 | ⋆₁, hdg_1 | ⋆₂, hdg_2 | hdg
+# ⋆₀⁻¹, invhdg_0 | ⋆₁⁻¹, invhdg_1 | ⋆₂⁻¹, invhdg_2 | hdg
+# Δ₀, lapl_0 | Δ₁, lapl_1 | Δ₂, lapl_2 | lapl
+# δ₁, codif_1 | δ₂, codif_2 | codif
+# avg₀₁, avg_01 | avg
+# -, neg | mag, norm
+
+CANON_NAMES = Dict{Symbol, Symbol}(
+  # Partial time derivative
+  :∂ₜ => PARTIAL_T,
```
lukem12345, line 160: Why is this not given a `const UNICODE_PARTIALT`

GeorgeR227:
I've skimmed through the rest of the code and it seems like really the only other function that should use canon names, in DiagrammaticEquations, would be the `find_chains` code. So in this case, we can just check that the canon name of an operator matches the canon name of an operator in a black/whitelist.

Everything else seems to just be editing names or copying over raw names and so don't need this.

src/deca/deca_op_names.jl
```diff
@@ -0,0 +1,254 @@
...
+const UNICODE_LAPLACE_1 = :Δ₁
+const UNICODE_LAPLACE_2 = :Δ₂
+
+const UNICODE_CODIF_1 = :δ₁
+const UNICODE_CODIF_2 = :δ₂
+
+const UNICODE_AVG_01 = :avg₀₁
+
+# Op2 names
+const UNICODE_WEDGE_00 = :∧₀₀
```
lukem12345, line 111: I would group these blocks of operators such that all forms of wedge product (i.e.g including ASCII) are grouped together.

src/deca/deca_op_names.jl
GeorgeR227: Because the unicode consts are only needed for overloading and this should never be overloaded.

GeorgeR227: Really, unicode names should just be aliases for ascii names but I have the unicode consts for nicer function naming in overloading, so for example `op is ∧ => ∧₀₁ but if is wdg => wdg_01`.

lukem12345:
https://github.com/AlgebraicJulia/Decapodes.jl/pull/142 is a spiritual predecessor of this feature. The main distinctions are that the old PR "canonicalized" on Unicode, and this PR ASCII, and the old PR added a subroutine that edits the ACSet to use the canonicalized names.

lukem12345:
We note that `incident(decapode, :L_1, :op1)` needs special treatment if it is to be inter-operable with canonical representations. Of course, features such as `incident(decapode, :L, :op1)` (the intention being to get all Lie derivatives, typed or untyped, Unicode or ASCII) require more engineering anyway.

lukem12345:
As discussed off-line, we should:

- [ ] Have a way of getting the Unicode of representation of an operator. (Currently partially handled by the Dict from Decapodes.jl PR 142)
- [x] '' ASCII ''. Currently handled by this PR.
- [ ] We want to get the set of all aliases. This information is stored implicitly now.

PR #60, opened 2024-07-29
Higher level incident call

GeorgeR227:
This PR adds a higher level `collected_incident` function that is meant to serve as a replacement to the many reduce-vcats we tend to use when trying to call incident on multiple things at once. Since results from this function are always combined into a single vector of unique results, as opposed to `incident` which may return a vector of vector of results, care must be taken when calling this function into different tables or when calling with multiple inputs.

This function is meant to be generic to work well with the idea that DiagrammaticEquations should have abstracted algorithms but I've also added specific helper functions for Decapodes that may be useful.

### DIFF
--- a/src/deca/deca_query.jl
+++ b/src/deca/deca_query.jl
@@ -1,0 +1,25 @@
+using DiagrammaticEquations
+using ACSets
+
+export is_var_target, is_var_source, get_variable_parents, get_next_op1s, get_next_op2s
+
+function is_var_target(d::SummationDecapode, var::Int) 
+  return !isempty(collected_incident(d, var, [:tgt, :res, :sum]))
+end
+
+function is_var_source(d::SummationDecapode, var::Int) 
+  return !isempty(collected_incident(d, var, [:src, :proj1, :proj2, :summand]))
+end
+
+function get_variable_parents(d::SummationDecapode, var::Int)
+  return collected_incident(d, var, [:tgt, :res, :res, [:summation, :sum]], [:src, :proj1, :proj2, :summand])
+end
+
+function get_next_op1s(d::SummationDecapode, var::Int)
+  collected_incident(d, var, [:src])
+end
+
+function get_next_op2s(d::SummationDecapode, var::Int)
+  collected_incident(d, var, [:proj1, :proj2])
+end
+
--- a/test/deca_query.jl
+++ b/test/deca_query.jl
@@ -1,0 +1,111 @@
+using Test
+using DiagrammaticEquations
+using DiagrammaticEquations.Deca
+using ACSets
+
+function array_contains_same(test, expected)
+    sort(test) == sort(expected)
+end
+
+get_index_from_name(d::SummationDecapode, varname::Symbol) = only(incident(d, varname, :name))
+
+@testset "Check sources and targets" begin
+  singleton_deca = @decapode begin
+    V::infer
+  end
+
+  @test !is_var_source(singleton_deca, 1)
+  @test !is_var_target(singleton_deca, 1)
+
+
+  path_op1_deca = @decapode begin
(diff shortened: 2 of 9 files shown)


lukem12345:
Some calls may be better off as `@generated` functions. We should talk about what it would take for this to be in ACSets main at Wednesday's Catlab Developer's meeting.

GeorgeR227:
I'd rather not be writing `@generated` calls since we've seen that they don't have their intended effects in ACSets anyway and that'd severely complicate the code.

Having this in ACSets main sounds fine but it leads me to wonder what exactly is supposed to be the difference between ACSets and DiagrammaticEquations.

lukem12345:
To ACSets.jl belong data structures and algorithms over generic (not only compute-graph like) schemas. Partitioning the code as of the most recent commit, `src/query.jl` and `test/query.jl` are ACSets.jl-like.

The "query" related code is quite similar to the chained-subparts code, which Sean contributed to ACSets.jl, in that case using CompTime. A recent draft PR demonstrates how using `@generated` to replace that feature allows the Julia compiler to e.g. eliminate intermediate allocations from certain loops.

Sean's PR provides syntactic sugar for nested calls to `incident`. This current PR for adjacent. The code written for the current PR can only be improved by consulting with those sister ones.

test/deca_query.jl
```diff
@@ -0,0 +1,111 @@
+using Test
+using DiagrammaticEquations
+using DiagrammaticEquations.Deca
+using ACSets
+
+function array_contains_same(test, expected)
+    sort(test) == sort(expected)
```
jpfairbanks, line 7: so you want to handle multiplicity? Could replace with `setequal(a,b) = Set(a) == Set(b)`

src/deca/deca_query.jl
```diff
@@ -0,0 +1,25 @@
...
+function is_var_target(d::SummationDecapode, var::Int) 
+  return !isempty(collected_incident(d, var, [:tgt, :res, :sum]))
+end
+
+function is_var_source(d::SummationDecapode, var::Int) 
+  return !isempty(collected_incident(d, var, [:src, :proj1, :proj2, :summand]))
+end
+
+function get_variable_parents(d::SummationDecapode, var::Int)
+  return collected_incident(d, var, [:tgt, :res, :res, [:summation, :sum]], [:src, :proj1, :proj2, :summand])
```
jpfairbanks, line 15: I have no idea what this is supposed to do, is this supposed to be a conjucnction of disjunctions type query?

jpfairbanks (changes requested):
@GeorgeR227, could we replace some of this with calls to `querry` from Catlab? 

https://github.com/AlgebraicJulia/Catlab.jl/blob/d336c7bb6b2f7b7e0e105553918c4caccdd99d6f/src/wiring_diagrams/Algebras.jl#L158

It lets you use the relation macro as a declarative DSL for queries. 

Since most queries are a chain of subpart and incident calls, you just write those as simple relations.

If you want to find all the variables that are computed by a direct circularity.
```julia
q = @relation begin
  Op1(src=v, tgt=u)
  Op1(src=u, tgy=v)
end
d = SummationDecapode(...)
query(d, q)
```

This would match patterns like `u = f(v); v = g(u)`. But then you can also do complex patterns like:

```julia
@relation begin
  Op1(src=v, tgt=u)
  Op2(proj1=v, proj2=u, res=x)
end
```

This is a pattern of a triangular dependency where what looks like an op2 on v,u is actually a unary operation on v, because u is derived from v. 

```julia
u = f(v)
x = g(v,u)
```

lukem12345:
This PR was good for testing out ideas that are now captured by https://github.com/AlgebraicJulia/ACSets.jl/pull/163

PR #62, opened 2024-08-05
Metatheory backend POC

olynch:


roe/README.md
```diff
@@ -0,0 +1,133 @@
...
+end
+```
+
+The type parameter `T` could be instantiated with `BasicSymbolic{<:DECVar}` or `Var`, depending on whether we are working with SymbolicUtils or Metatheory.
+
+So, for instance, the Klausmeier model might look like
+
+```julia
+function klausmeier(roe::Roe)
+  @vars roe n::DualForm0 w::DualForm0 dX::Form1 a::Constant{DualForm0} ν::Constant{DualForm0}
```
jpfairbanks, line 88: Should these be replaced by `n::Form(0,Dual)` or something? or does `@vars` handle legacy decapodes names and replace them?

roe/README.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+Namespacing is achieved by moving the roe into a namespace before passing it into submodels. So, for instance, to make a model with two Klausmeier submodels that share the same `m`, we could do:
+
+```julia
+function double_klausmeier(roe::Roe)
+  klausmeier(namespaced(roe, :k1))
+  klausmeier(namespaced(roe, :k2))
+
+  @eq roe (roe.k1.m == roe.k2.m)
```
jpfairbanks, line 103: having them share the same dX and w would make sense because that would give you two species of plant sharing the same water supply. Although at thought point, you would want to break the models down into smaller pieces and reassemble it with only one water supply.

This is how you would it in current Decapodes. You have to break out any factor that gets superposition into its own models so that colimits the category of diagrams do the right thing. 

```julia
# See Klausmeier Equation 2.a
Hydrodynamics = @decapode begin
  (w, consumption)::DualForm0
  dX::Form1
  (a,ν)::Constant
  ∂ₜ(w) == a - w - consumption + ν * L(dX, w)
end

# See Klausmeier Equation 2.b
Phytodynamics = @decapode begin
  (n,w)::DualForm0
  m::Constant
  consumption = w * n^2
  ∂ₜ(n) == consumption - m*n + Δ(n)
end

Superposition = @decapde begin
  total = part1 + part2
end

compose_klausmeier = @relation () begin
  phyto(N1, W, consumption1)
  phyto(N2, W, consumption2)
  superposition(consumption, consumption1, consumption2)
  hydro(consumption, W)
end

klausmeier_cospan = oapply(compose_klausmeier,
                           [Open(Phytodynamics, [:consumption, :w]), Open(Phytodynamics, [:consumption, :w])
                            Open(Hydrodynamics, [:total, :part1, :part2]),
                            Open(Hydrodynamics, [:consumption, :w])])
Klausmeier = apex(klausmeier_cospan)
```

I guess you could do it in Roe by replacing the definition of a term with a sum? That seems like it would cause you to flush the egraph and resaturate, because you removed an equation.

roe/src/roe.jl
```diff
@@ -0,0 +1,215 @@
...
+export SortError
+
+Base.get(lookup::Dict{TA,Any}, key::TA) = lookup[key]
+export get
+
+# DECVar{s}
+struct OfSort{s} <: Number end
+export OfSort
+
+struct SortMetadata end
```
jpfairbanks, line 51: Was this for using metadata for the types, but actually unnecessary because you figured out how to use `SymbolicUtils.Sym{OfSort{sort}}(name)` to put the actual types in there?

jpfairbanks (changes requested):
This is coming along nicely. It looks like for the Symbolics version, the big feature that we are missing is an analogue of extraction. It would be good if we could reuse the built in compiler for SymbolicUtils.  https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/src/code.jl

Specifically, we should be able to use `toexpr` to generate functions that evaluate the tangent variables. https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/2ef075e4274746d9c4b6c3215f385867b0c85edf/src/code.jl#L348

PR #64, opened 2024-08-14
Interop with SymbolicUtils

olynch:
This is a more-or-less feature-complete conversion between DecaExpr and an equivalent data structure built on SymbolicUtils, but it requires some testing/naming bikeshedding.


lukem12345:
@GeorgeR227 Looks like this should get rebased onto https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/54 ; there is apparently some redundancy in dicts for typing/ untyping symbols.

src/ThDEC.jl
```diff
@@ -0,0 +1,270 @@
...
+    @match s begin
+        Scalar() => throw(SortError("Cannot take Hodge star of a scalar"))
+        Form(i, isdual) => Form(2 - i, !isdual)
+    end
+end
+
+function Base.nameof(::typeof(★), s)
+    inv = isdual(s) ? "⁻¹" : ""
+    Symbol("★$(as_sub(isdual(s) ? 2 - dim(s) : dim(s)))$(inv)")
+end
```
GeorgeR227, line 151: I think we need a deeper consideration of the dimension that our physics in occurring in here. I've noticed that the current code is mostly hard-coded for 2D and in the case of the Hodge star, the resulting operator would be wrong if we were working in 1D.

I can say that the way our current inference/overloading rules work is by grouping them into contexts based on the dimension, so we have separate rules to apply in the 1D case and in the 2D case. The way the dimension is figured out is by the user explicitly passing us that information, since some physics can work without modification in multiple dimensions, like the Heat equation.

GeorgeR227:
> @GeorgeR227 Looks like this should get rebased onto #54 ; there is apparently some redundancy in dicts for typing/ untyping symbols.

I'm not sure because in this case `OPERATOR_LOOKUP` is taking typed names down to generic names. In that PR we take those typed names down to a single representative which is still typed. It's true that they're similar but their purposes feel different.

lukem12345:
Yes the rebase is for purposes of grouping together code that accomplishes similar tasks.

src/ThDEC.jl
quffaro: I think one way to work around this is to store the Space as a parameter in the Form sort. Space may be a wrapper for the dimension of Space, but can be embiggened with other attributes down the line, or maybe even a pointer to the mesh on which its defined, with the provision we have to update our signature checking.

I think this makes philosophical sense. n-forms are the same in any (n+k)-dimensional space, but since they are also determined by duality, their dual varies according to the dimensionality. So this means every n-form is determined by a different dual per dimension. Tagging this dimension `Form(dim, isdual, space)` allows us to track this.

@lukem12345 what do you think about this philosophically and practically?

lukem12345: Philosophically, here are some probing questions: How do you want to distinguish between the Heat equation in 1D, 2D, and 3D? What do morphisms between diagrams (of possibly differing dimension) look like?

Practically: the entire PR seems to be hard-coded for the 2D Discrete Exterior Calculus over a double de Rham complex. I believe more consideration needs to be taken as to how we reconcile having 3 similar algebraic theories in the same package. Is there any advantage to besides code reduction? What is the scale of that code reduction? And is the generalization worth the mental overhead? Are we going to be auto-generating rules by doing index arithmetic on these parameters or not? Do we want it to be clear when we are programming in a 1D environment vs a 3D environment? How does this impact representations with respect to the multi-dimensional multiphysics project? Do we want to represent the various DECs over a single de Rham complex?

olynch: This all makes sense. This was supposed to be a first stab at what the API would look like. I think the next step is to add the space to the types, and part of the information about the space will be its dimension.

olynch: When we convert a DecaExpr to a DecaSymbolic, you should have to also provide a mapping from space symbol to dimension.

olynch: What would you all think of something like:

```julia
struct Space
    name::Symbol
    dim::Int
end

@data Sort begin
    Scalar()
    Form(dim::Int, isdual::Bool, space::Space)
    VField(isdual::Bool, space::Space)
end
```

quffaro: @jpfairbanks and I discussed this exact suggestion earlier today. I'd like to implement it, but I think @lukem12345 's practical questions still remain open. @lukem12345 @GeorgeR227 for clarification, should these questions + considerations should be resolved in this PR?

lukem12345: Nothing is immediately keeping this PR open as long as tests pass.

src/decasymbolic.jl
```diff
@@ -0,0 +1,225 @@
...
+            v::BasicSymbolic{T}
+        ) where {T<:DECType}
+            # convert the DECType to ThDEC to type check
+            s = ThDEC.$unop(Sort(T))
+            # the resulting type is converted back to DECType
+            # the resulting term has the operation has its head and `v` as its args.
+            SymbolicUtils.Term{Number(s)}(ThDEC.$unop, [v])
+        end
+    end
+end
```
jpfairbanks, line 90: this loop should be a function that users can eval to register new symbolic functions. We almost always need at least 1 arbitrary functions that users should be able to register with a syntax like

```julia
@register f(x::PrimalForm{1})::PrimalForm{1}
```

It would also be cool if they could give the implementation there too. 

```julia
@register f(x::PrimalForm{1})::PrimalForm{1} = map(x) do val
  val  ^ 2
end
```

src/decasymbolic.jl
```diff
@@ -0,0 +1,225 @@
...
+        @nospecialize
+        function ThDEC.$unop(
+            v::BasicSymbolic{T}
+        ) where {T<:DECType}
+            # convert the DECType to ThDEC to type check
+            s = ThDEC.$unop(Sort(T))
+            # the resulting type is converted back to DECType
+            # the resulting term has the operation has its head and `v` as its args.
+            SymbolicUtils.Term{Number(s)}(ThDEC.$unop, [v])
+        end
```
jpfairbanks, line 88: why don't we have to export these like the binops?

jpfairbanks (changes requested):

jpfairbanks:
Instead of creating julia functions for the operators of the DEC, could we just create SymbolicUtils function types? 

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/df8e84d31e60f4dc70f0df2ad88ede91f6f3fd27/src/types.jl#L913

That is what you get when you do:

```julia
julia> @syms f(x)::Complex
(f(::Number)::Complex,)
```

src/decasymbolic.jl
```diff
@@ -0,0 +1,225 @@
...
+"""
+function SymbolicUtils.BasicSymbolic(context::Dict{Symbol,Sort}, t::decapodes.Term)
+    @match t begin
+        Var(name) => SymbolicUtils.Sym{Number(context[name])}(name)
+        Lit(v) => Meta.parse(string(v)) # TODO no YOLO
+        # see heat_eq test: eqs had AppCirc1, but this returns
+        # App1(f, App1(...)
+        AppCirc1(fs, arg) => foldr(
+            # panics with constants like :k
+            # see test/language.jl
```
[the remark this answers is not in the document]
lukem12345: Also fits into best practice of "being liberal in what we accept" https://en.wikipedia.org/wiki/Robustness_principle

src/ThDEC.jl
```diff
@@ -0,0 +1,326 @@
...
+export Space
+
+dim(s::Space) = s.dim
+Base.nameof(s::Space) = s.name
+
+struct SpaceLookup
+    default::Space
+    named::Dict{Symbol,Space}
+end
```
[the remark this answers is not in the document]
jpfairbanks: Realized this needs to be default name and dimension

src/decasymbolic.jl
quffaro: Tests currently pass with exception to `klausmeier`. To get it working, @jpfairbanks suggested a `@register` macro to add function terms. However I'm not sure if I'm understanding completely, so I'm seeking clarification. 

Currently (in this PR), for a given operator we define two methods:
1. a  Julia function over `ThDEC` for checking its type. They are aggregated in a OPERATOR_LOOKUP const. This step is analogous to SymbolicUtil's `promote_symtype`
2. a Julia function over `DecaSymbolic`, our type for the SymbolicUtils implementation of ThDEC, which calls (1) for validation

we use the OPERATOR_LOOKUP to also check if a model has admissible terms (e.g.,  Δ), so we want the ability to append the OPERATOR_LOOKUP by `@register`ing new functions.

```{julia}
@register foo(x::Scalar)::Scalar = x + 1
# codegens
foo(s::Scalar) = Scalar
foo(x::BasicSymbolic{T<:DECType}) = begin
    s = foo(Sort(T)) # TYPE CHECKING
    Sym{FnType{Tuple{Number(s)}, Real}(foo)
end
```

Another example comes with a hazard: `PrimalForm{T}` is a parameterized type, but `PrimalForm::Sort` is a function. If we want to continue typechecking with `ThDEC` types, do our current way of checking signatures, we need to make sure codegen for typechecking supports the engineering task to do this.

```{julia}
@register foo(x::PrimalFormT{1})::PrimalFormT{1}
# codegens
foo(x::Form) = @match x begin
    PrimalForm(1) => PrimalForm(1)
    _ => error
end
```

## The steps:

Suppose we wanted to add `foo` to the OPERATOR_LOOKUP dictionary. Notice `PrimalFormT{1}` is the DecaSymbolic type, which accepts parameters.
```{julia}
@register foo(x::PrimalFormT{1})::PrimalFormT{1} = map(x) do val
  val  ^ 2
end
```
We need to define its (1) type-checking (`src/ThDEC`), (2) its DecaSymbolic implementaton (`src/decasymbolic`), and possibly (3) its implementation.

We define the implementation of a function
```{julia}
func(x::PrimalFormT{1})::PrimalFormT{1} = map(x) do val; val ^ 2 end
```
:question: Where does this live? Does this act on BasicSymbolic terms?

Next, we define it symbolically:
```{julia}
symfunc = Sym{FnType{Args, Real}}(ThDEC.func)
```

Then, we add it to OPERATOR_LOOKUP (possible `haskey` validations omitted)
```
ThDEC.OPERATOR_LOOKUP[op] = func
```
:question: Currently OPERATOR_LOOKUP is typed as `Dict{Symbol, Function}`. Is the intent to change `Dict{Symbol, SymbolicFunction}` and do typechecking by replacing the # TYPE CHECKING (above) line with `promote_symtype`?

lukem12345:
An implication of Decapode type-inference that seems baked-into this PR is that it is assumed that the choice of primal/ dual is completely inferable based on the types of the parents. However certain operations such as the musical isomorphisms or wedge products can either return a primal or dual form depending on the discrete operator chosen. Note also that for any Decapode, there are multiple means of specifying whether forms are primal or dual. Performing type inference via breadth-first or depth-first traversal at parse-time biases the system towards only representing certain discretizations. In certain cases, choosing a particular discretization for a certain variable or operation may constrain successor operations to certain discretizations, for which CombinatorialSpaces may not yet offer a method. And so a satisfiable model may not be returned, although one may exist.

jpfairbanks:
We can always do dispatch on return type to let you write sharp(Dual, x)

lukem12345:
Yes but that is still local

jpfairbanks:
I'm moving a comment from @quffaro from inline reply to top level comment so that it stays here after we change the code and mark that thread resolved:

Tests currently pass with exception to `klausmeier`. To get it working, @jpfairbanks suggested a `@register` macro to add function terms. However I'm not sure if I'm understanding completely, so I'm seeking clarification. 

Currently (in this PR), for a given operator we define two methods:
1. a  Julia function over `ThDEC` for checking its type. They are aggregated in a OPERATOR_LOOKUP const. This step is analogous to SymbolicUtil's `promote_symtype`
2. a Julia function over `DecaSymbolic`, our type for the SymbolicUtils implementation of ThDEC, which calls (1) for validation

we use the OPERATOR_LOOKUP to also check if a model has admissible terms (e.g.,  Δ), so we want the ability to append the OPERATOR_LOOKUP by `@register`ing new functions.

```{julia}
@register foo(x::Scalar)::Scalar = x + 1
# codegens
foo(s::Scalar) = Scalar
foo(x::BasicSymbolic{T<:DECType}) = begin
    s = foo(Sort(T)) # TYPE CHECKING
    Sym{FnType{Tuple{Number(s)}, Real}(foo)
end
```

Another example comes with a hazard: `PrimalForm{T}` is a parameterized type, but `PrimalForm::Sort` is a function. If we want to continue typechecking with `ThDEC` types, do our current way of checking signatures, we need to make sure codegen for typechecking supports the engineering task to do this.

```{julia}
@register foo(x::PrimalFormT{1})::PrimalFormT{1}
# codegens
foo(x::Form) = @match x begin
    PrimalForm(1) => PrimalForm(1)
    _ => error
end
```

## The steps:

Suppose we wanted to add `foo` to the OPERATOR_LOOKUP dictionary. Notice `PrimalFormT{1}` is the DecaSymbolic type, which accepts parameters.
```{julia}
@register foo(x::PrimalFormT{1})::PrimalFormT{1} = map(x) do val
  val  ^ 2
end
```
We need to define its (1) type-checking (`src/ThDEC`), (2) its DecaSymbolic implementaton (`src/decasymbolic`), and possibly (3) its implementation.

We define the implementation of a function
```{julia}
func(x::PrimalFormT{1})::PrimalFormT{1} = map(x) do val; val ^ 2 end
```
:question: Where does this live? Does this act on BasicSymbolic terms?

Next, we define it symbolically:
```{julia}
symfunc = Sym{FnType{Args, Real}}(ThDEC.func)
```

Then, we add it to OPERATOR_LOOKUP (possible `haskey` validations omitted)
```
ThDEC.OPERATOR_LOOKUP[op] = func
```
:question: Currently OPERATOR_LOOKUP is typed as `Dict{Symbol, Function}`. Is the intent to change `Dict{Symbol, SymbolicFunction}` and do typechecking by replacing the # TYPE CHECKING (above) line with `promote_symtype`?

jpfairbanks:
> We define the implementation of a function
> 
> ```
> func(x::PrimalFormT{1})::PrimalFormT{1} = map(x) do val; val ^ 2 end
> ```
> 
> ❓ Where does this live? Does this act on BasicSymbolic terms?

I am rethinking my idea to define the implementation here. The implementations will often need to refer to parts of the mesh, including vertex coordinates or volumes of a simplex. So we should reserve this syntax for defining a new function symbol and a definition in terms of more primitive functions that could get turned into a rewrite. So that you could register a function like 

```julia
f, r = @register f(v::PrimalFormT{1},x::PrimalFormT{0})::PrimalFormT{1} = d(Δ(x)) + (v  ∧ x)
```
and have that expand out to

```julia
f, = @syms f(v::PrimalFormT{1},x::PrimalFormT{0})::PrimalFormT{1}
r = @rule f(~v::PrimalFormT{1},~x::PrimalFormT{0})::PrimalFormT{1} => d(Δ(~x)) + (~v  ∧ ~x)
```

> 
> Next, we define it symbolically:
> 
> ```
> symfunc = Sym{FnType{Args, Real}}(ThDEC.func)
> ```
> 
> Then, we add it to OPERATOR_LOOKUP (possible `haskey` validations omitted)
> 
> ```
> ThDEC.OPERATOR_LOOKUP[op] = func
> ```
> 
> ❓ Currently OPERATOR_LOOKUP is typed as `Dict{Symbol, Function}`. Is the intent to change `Dict{Symbol, SymbolicFunction}` and do typechecking by replacing the # TYPE CHECKING (above) line with `promote_symtype`?

I think that we can create a function for each DEC operator, then create a symbolic function with that function as the head. 

```julia
function Δ end
Δ, = @syms Δ(x::PrimalFormT{0})::PrimalFormT{0}
```

If the function symbol we are creating needs to do dispatch, then it should overload `promote_symtype` to hook that logic into SymbolicUtils the way that it is expected.

lukem12345:
https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/issues/35

I think using a defintion operator := like that mentioned in this old issue would be nice.

src/ThDEC.jl
```diff
@@ -5,6 +5,31 @@ using StructEquality
...
+    end
+    result = quote end
+    foreach(aliases) do alias
+        push!(result.args,
+            quote
+                function $(esc(alias))(s...)
+                    $(esc(rep))(s...) 
+                end
+                export $(esc(alias))
+                Base.nameof(::typeof($alias), s) = nameof($rep, s)
```
jpfairbanks, line 27: Do we want the nameof to be the symbol for the alias or the rep? This might break roundtriping because when you print something that used an alias, it will be shown as the rep.

src/ThDEC.jl
```diff
@@ -180,20 +205,23 @@ function d(s::Sort)
     end
 end
 
-function Base.nameof(::typeof(d), s)
-    Symbol("d$(as_sub(dim(s)))")
-end
+@alias d, (d₀, d₁)
```
jpfairbanks, line 208: maybe using an oriented syntax like 

`@alias (d₀, d₁) => d` will help people remember which is the alias name and which is the alias result.

src/ThDEC.jl
```diff
@@ -180,20 +205,23 @@ function d(s::Sort)
...
+
+Base.nameof(::typeof(d), s) = Symbol("d$(as_sub(dim(s)))")
 
 @nospecialize
-function ★(s::Sort)
+function ⋆(s::Sort)
```
jpfairbanks, line 213: oh I didn't know about \bigstar. We should use that for hodge star going forward.

src/decasymbolic.jl
```diff
@@ -85,8 +74,13 @@ Number(f::Form) = FormT{dim(f),isdual(f), nameof(space(f)), dim(space(f))}
 
 Number(v::VField) = VFieldT{isdual(v), nameof(space(v)), dim(space(v))}
 
+# HERE WE DEFINE THE SYMBOLICUTILS
+
 # for every unary operator in our theory, take a BasicSymbolic type, convert its type parameter to a Sort in our theory, and return a term
-unop_dec = [:∂ₜ, :d, :★, :♯, :♭, :-]
+unop_dec = [:∂ₜ, :d, :d₀, :d₁
+            , :⋆, :⋆₀, :⋆₁, :⋆₂, :⋆₀⁻¹, :⋆₁⁻¹, :⋆₂⁻¹
+            , :♯, :♭, :-]
```
jpfairbanks, line 82: I don't like commas at the start of the next line. I get that it is valid Julia, but it looks weird.

src/decasymbolic.jl
```diff
@@ -99,6 +93,8 @@ for unop in unop_dec
     end
 end
 
+# BasicSymbolic{FnType{Tuple{PrimalFormT{0}}}, PrimalFormT{0}}
```
jpfairbanks, line 96: can delete comment

src/decasymbolic.jl
```diff
@@ -211,6 +207,8 @@ Example:
 ```
 """
 function SymbolicUtils.BasicSymbolic(context::Dict{Symbol,Sort}, t::decapodes.Term)
+    # user must import symbols into scope
+    ! = (f -> getfield(Main, f))
```
jpfairbanks, line 211: is main always the right module to evaluate this in? I think that we can pass the module in as an argument and have it default to Main. You might want to have two models in the same program that live in different modules.

```
function SymbolicUtils.BasicSymbolic(context::Dict{Symbol,Sort}, t::decapodes.Term, module::Module=Main)
```

should work for a signature.

src/decasymbolic.jl
```diff
@@ -276,6 +275,66 @@ function SummationDecapode(e::DecaSymbolic)
...
+              @nospecialize
+              function $(esc(f))($(newargs...)) where $(constraints...)
+                  s = $(esc(f))($(innerargs...))
+                  SymbolicUtils.Term{Number(s)}($(esc(f)), [$(getindex.(binding, 1)...)])
+              end
+          end)
+
+    return result
+end
+export @register
```
jpfairbanks, line 338: Can we use this in all the current unops and binops definitions? It would be good to dogfood the macro to make sure it handles all the cases that we need.

jpfairbanks (changes requested):
Looking closer. I put some inline comments on there.

src/decasymbolic.jl
quffaro: Yeah I can one line this. This style is something I picked up from SQL

src/ThDEC.jl
quffaro: Good call, I'll update

src/decasymbolic.jl
jpfairbanks: You can have a struct with a field that contains a module as a value. So if you just add the module of resolution to the struct as a field and have a macro that uses @__module__ and captures it for future use that is fine. We also only need to resolve which module when we are converting between DecaExpr/ACset and DecaSymbolic. Once we are in DecaSymbolic world, the only module of resolution should the current one. And we shouldn't need to call getfield on a module because we aren't converting symbols to functions, we just have references to the functions we need in the current scope.

src/symbolictheoryutils.jl
```diff
@@ -6,6 +6,7 @@ import SymbolicUtils: promote_symtype
 function rules end
 export rules
 
+# TODO: Probable piracy here
 function promote_symtype(f::ComposedFunction, args)
```
jpfairbanks, line 10: We should upstream this method https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/650

test/decasymbolic.jl
```diff
@@ -0,0 +1,281 @@
...
+  @test nameof(d, symtype(u)) == Symbol("d₀")
+  @test_broken nameof(d, symtype(η)) == Symbol("dual_d₂")
+
+  @test_broken nameof(Δ, symtype(u)) == Symbol("Δ₀")
+  @test_broken nameof(Δ, symtype(ω)) == Symbol("Δ₁")
+
+  @test nameof(★, symtype(u)) == Symbol("★₀")
+  @test nameof(★, symtype(ω)) == Symbol("★₁")
+  @test nameof(★, symtype(η)) == Symbol("★₀⁻¹")
+end
```
GeorgeR227, line 114: @lukem12345 I noticed while going through the `Base.nameof` functionality that support for names of operators with dual inputs is not well supported. Could you advise on how we should be naming these operators to differentiate them from our common primal-primal operators?

GeorgeR227: I think most notable here is the wedge and the Laplacian.

jpfairbanks: I think that either dual_ or \~ would be appropriate.

PR #67, opened 2024-08-22
Add type rules for vectorfields as symbols

lukem12345:


src/acset.jl
```diff
@@ -427,12 +430,12 @@ function safe_modifytype!(d::SummationDecapode, var_idx::Int, new_type::Symbol)
 end
 
 """
-    filterfor_forms(types::AbstractVector{Symbol})
+    filterfor_ec_types(types::AbstractVector{Symbol})
```
jpfairbanks, line 433: should we have both? if we rename, have to bump minor version number.

test/language.jl
```diff
@@ -831,6 +832,38 @@ end
...
+
+      M == ♯(I)
+      N == ♯(J)
+      O == ♭(K)
+      P == ♭(L)
+    end
+    infer_types!(d)
+
+    # TODO: Update this as more sharps and flats are released.
+    names_types_expected = Set([(:A, :Form1), (:B, :DualForm1), (:C, :PVF), (:D, :DVF),
```
jpfairbanks, line 861: This should be a Dict{Symbol, Symbol} right? each variable needs a unique type.

jpfairbanks (changes requested):

test/language.jl
lukem12345: That would rely on the property of names in a Decapode being unique

src/acset.jl
lukem12345: This function is not exported

test/language.jl
jpfairbanks: Good point.

jpfairbanks (approved):

PR #68, opened 2024-08-23
amending pipeline

quffaro:
This updates the pipeline.yml file

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,6 @@
 env:
-  JULIA_VERSION: "1.9.3"
+  JULIA_VERSION: "1.10.2"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,5 +1,6 @@
 env:
-  JULIA_VERSION: "1.9.3"
+  JULIA_VERSION: "1.10.2"
+  JULIA_DEPOT_PATH: "$DEPOT"
```
jpfairbanks, line 3: Does this work? It pulls the path from the environment of the agent?

quffaro: yes, this is initiated by an environment hook which the buildkite-agent is pointed to when we spawn them. Now to change the depot we just need to amend the hook and restart the agents.

PR #69, opened 2024-08-23
Expression level rewriting

GeorgeR227:
This PR is meant to create a back-and-forth between ACSets, which excel at information propagation (typing, overloading, etc.) and Symbolics, which excel at expression rewriting (open_operators, optimizations, etc.). 

This should remove the need for our graph rewriting functions and make the workflow to add new rewrites into the system much more faster/cleaner.


src/graph_traversal.jl
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+  end
+
+  @assert length(op_order) == number_of_ops(d)
+
+  op_order
+end
+
+function number_of_ops(d::SummationDecapode)
```
jpfairbanks, line 57: nops?

src/graph_traversal.jl
```diff
@@ -0,0 +1,72 @@
...
+function start_nodes(d::SummationDecapode)
+  return vcat(infer_states(d), incident(d, :Literal, :type))
+end
+
+function retrieve_name(d::SummationDecapode, tsr::TraversalNode)
+  @match tsr.name begin
+    :Op1 => d[tsr.index, :op1]
+    :Op2 => d[tsr.index, :op2]
+    :Σ => :+
+    _ => error("$(tsr.name) is not a valid table for names")
```
jpfairbanks, line 70: "$(tsr.name) is a table without names"

test/graph_traversal.jl
```diff
@@ -0,0 +1,64 @@
+using DiagrammaticEquations
+using ACSets
+using MLStyle
+using Test
+
+function is_correct_length(d::SummationDecapode, result)
+  return length(result) == number_of_ops(d)
+end
```
jpfairbanks, line 8: We can also check that when you iterate over the edges in the decapode, we never go from a later vertex to an earlier vertex.

jpfairbanks (changes requested):
The topo sort stuff is looking good. The symbolics integration should target the existing symbolics integration branch.

lukem12345:
When roundtripping to the SummationDecapode is complete, it would be good to test that the results in `test/openoperators.jl` give the same results up to renaming etc

src/graph_traversal.jl
```diff
@@ -0,0 +1,72 @@
+using DiagrammaticEquations
+using ACSets
+
+export TraversalNode, topological_sort_edges, number_of_ops, retrieve_name
+
+struct TraversalNode{T}
+  index::Int
+  name::T
+end
```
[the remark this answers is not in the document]
jpfairbanks: Ok, the unitype design that you did, coupled with an `Enum` that captures the available symbols would be good. https://docs.julialang.org/en/v1/base/base/#Base.Enums.@enum and https://thautwarm.github.io/MLStyle.jl/latest/syntax/pattern.html#support-pattern-matching-for-julia-enums could work together to make it easy to write the generic code.

GeorgeR227:
I've rebased this branch to work off of PR #64 since the code here will be used for the rewriting using only `Sort`.

src/sym_rewrite.jl
lukem12345, line 1: Why is this a source file? @GeorgeR227 @quffaro

GeorgeR227:
@lukem12345 We wanted to remove this check since it lead to a lot of special code to handle the final ACSet generation. I've removed it in commit https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/69/commits/5b84cc8a93097832e2a0fa492bba913f17904bba and added the appropriate tests to ensure it's proper functionality.

src/sym_rewrite.jl
GeorgeR227: This is just a scripting file to test out the pipeline, I've tagged it for removal before merging.

src/acset2symbolic.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+formed_deca_eqn(symeqn::SymbolicEquation{Symbolic}) = SymbolicUtils.Term{Number}(DECA_EQUALITY_SYMBOL, [symeqn.lhs, symeqn.rhs])
+
+function to_acset(d::SummationDecapode, sym_exprs)
+  outer_types = map([infer_states(d)..., infer_terminals(d)...]) do i
+    :($(d[i, :name])::$(d[i, :type]))
+  end
+
+  tangents = map(incident(d, DerivOp, :op1)) do op1
```
lukem12345, line 76: This variable is assigned but never used.

GeorgeR227: Removed it now

test/decasymbolic.jl
lukem12345, line 1: @quffaro Merge commit 367414d causes “malformed export statement” warnings in the “Operator definition” test set.

src/acset2symbolic.jl
```diff
@@ -0,0 +1,78 @@
...
+function symbolics_lookup(d::SummationDecapode)
+  Dict{Symbol, BasicSymbolic}(map(d[:name],d[:type]) do name,type
+    (name, decavar_to_symbolics(d, name, type))
+  end)
+end
+
+function decavar_to_symbolics(d::SummationDecapode, var_name::Symbol, var_type::Symbol; space = :I)
+  new_type = SymbolicUtils.symtype(Deca.DECQuantity, var_type, space)
+  SymbolicUtils.Sym{new_type}(var_name)
+end
```
GeorgeR227, line 20: This no longer needs to be passed `d`.

lukem12345: Go for it

GeorgeR227: Thanks, done

PR #70, opened 2024-08-24
Floyd-Warshall

lukem12345:
The Floyd-Warshall algorithm can be defined on the variables of a Decapode. This entails a topological ordering of those variables. A topological ordering of operations is given in turn.


lukem12345:
Commit d46cf86 added both docstrings and introduced a multiple dispatch scheme by subtyping `TraversalNode`. The MD scheme can be undone depending on design constraints I’m not aware of. This results in cleaner algorithm code

lukem12345:
It would be a fun project to investigate whether and how this can be framed in terms of an “Algebraic Path Problem”.[1,2]

[1] https://drops.dagstuhl.de/storage/00lipics/lipics-vol211-calco2021/LIPIcs.CALCO.2021.20/LIPIcs.CALCO.2021.20.pdf
[2] https://edge.edx.org/asset-v1:BerkeleyX+CS188-SU16+SU16+type@asset+block/kleene_algebras_path_problems.pdf

src/graph_traversal.jl
```diff
@@ -6,67 +6,99 @@ export TraversalNode, topological_sort_edges, number_of_ops, retrieve_name
...
-  visited_1 = falses(nparts(d, :Op1))
-  visited_2 = falses(nparts(d, :Op2))
-  visited_Σ = falses(nparts(d, :Σ))
+TraversalNode(i, d::SummationDecapode, ::Val{:Op2}) =
+  TraversalNode{Symbol}(i, d[i,:op2], [d[i,:proj1],d[i,:proj2]], d[i,:res])
 
-  # FIXME: this is a quadratic implementation of topological_sort inlined in here.
-  op_order = TraversalNode{Symbol}[]
+TraversalNode(i, d::SummationDecapode, ::Val{:Σ}) =
+  TraversalNode{Symbol}(i, :+, d[incident(d,i,:summation),:summand], d[i,:sum])
```
GeorgeR227, line 20: Instead of the dispatching being a seperate argument, maybe we can bake that dispatching into the struct type, so we have `TraversalNode{Op1}` instead of `TraversalNode{Symbol}` with a `name` field that is :Op1.

src/graph_traversal.jl
```diff
@@ -6,67 +6,99 @@ export TraversalNode, topological_sort_edges, number_of_ops, retrieve_name
...
+Interpret a:
+  - unary operation as a hyperedge of order (1,1) ,
+  - binary operation as a hyperedge of order (2,1) , and
+  - summation as a hyperedge of order (|summands|,1) .
+"""
+function hyper_edge_list(d::SummationDecapode)
+  [map(e -> TraversalNode(e, d, Val(:Op1)), parts(d, :Op1))...,
+   map(e -> TraversalNode(e, d, Val(:Op2)), parts(d, :Op2))...,
+   map(e -> TraversalNode(e, d, Val(:Σ  )), parts(d, :Σ  ))...]
+end
```
GeorgeR227, line 46: Have a function that returns all of the "edge" symbols, so in this case `Op1, Op2, Σ` and just iterate over those with the interior functionality. Can also do this with all the "vertex" symbols.

GeorgeR227:
I get your idea about make our hypergraph algorithms more generic. I think many people have wanted that and I do want to keep those parts of this PR. 

What I don't get is the decision to base everything on Floyd-Warshall. If F-W can be made into a generic hypergraph algorithm, then why can't topological sort? The implementation was almost there, it just needed the concept of the generic edge (which you implemented here) plus a bit of other functionality probably.

My main idea here is that the decision to use F-W makes this code less scalable/performant, less readable (since this is no longer the standard topological sort algorithm most people are familiar with) and less maintainable (because now we have graph algorithms which depend on other graph algorithms when there's no reason they have to).

I like the push towards going generic (even though that wasn't the explicit push of the branch this is based off) and I don't even mind the inclusion of F-W but it should not be used in the topological sort.

lukem12345:
We're on the same page

lukem12345:
@GeorgeR227 The most recent commit needs to use `git mv`. We shouldn't delete files and place their contents in new files.

GeorgeR227:
Yeah I can do that. Any thoughts on the generic `HyperGraph`?

lukem12345:
We should probably just switch to using an incidence matrix. Also accommodates arguments of arbitrary arity. In Decapodes then just "dispatch"/ switch/ pattern-match on arity and name of symbol. The Restricting operations to unary ops, binary ops, and sums is no longer is a necessary design constraint.

lukem12345:
Or just use this schema, migration or otherwise. Nodes are rows in the Var table. Edges are cached calls to
```julia
(dom = d[incident(d,i,:arg_op),:arg_var], cod = d[i,:res], op = function[i])
```.

```julia
@schema SchVariadicDecapode begin
  Var::Ob
  (Type, VarName, Operator)::AttrType(Symbol)
  type::Attr(Var, Type)
  name::Attr(Var, Name)

  (Op, Argument)::Ob
  (Function)::AttrType(Symbol)
  function::Attr(Op, Function)
  arg_var::Hom(Argument, Var)
  arg_op::Hom(Argument, Op)
  res::Hom(Op, Var)
end
```

lukem12345:
Since this does not use the Decapode ACSet explicitly, and rather uses a "reduces in polynomial-time" coding style to simplify downstream Decapodes code, this PR will be deleted without merging.

PR #71, opened 2024-08-30
Cm/symbolictypes

quffaro:


src/SymbolicUtilsInterop.jl
```diff
@@ -0,0 +1,154 @@
...
+        )
+        App1(f, x) => (!(f))(BasicSymbolic(context, x, __module__))
+        App2(f, x, y) => (!(f))(BasicSymbolic(context, x, __module__), BasicSymbolic(context, y, __module__))
+        Plus(xs) => +(BasicSymbolic.(Ref(context), xs, Ref(__module__))...)
+        Mult(xs) => *(BasicSymbolic.(Ref(context), xs, Ref(__module__))...)
+        Tan(x) => ∂ₜ(BasicSymbolic(context, x, __module__))
+    end
+end
+
+function SymbolicContext(d::decapodes.DecaExpr, __module__=@__MODULE__)
```
jpfairbanks, line 108: When you call this with 1 arguement, does this `@__MODULE__` resolves to SymbolicUtilsInterop or to Main?

src/SymbolicUtilsInterop.jl
```diff
@@ -0,0 +1,154 @@
...
+    # associates each var to its sort...
+    context = map(d.context) do j
+        @info j.var
+        j.var => j.var
```
jpfairbanks, line 112: should the right hand side of this pair be the sort?

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+    dim::Int
+    duality::Bool
+    space::Symbol
+    spacedim::Int
+end
+
+dim(fp::FormParams) = getproperty(fp, :dim)
+duality(fp::FormParams) = getproperty(fp, :duality)
+space(fp::FormParams) = getproperty(fp, :space)
+spacedim(fp::FormParams) = getproperty(fp, :spacedim)
```
jpfairbanks, line 34: unnecessary use of getproperty. These should just use fp.spacedim etc.

src/SymbolicUtilsInterop.jl
```diff
@@ -0,0 +1,154 @@
+module SymbolicUtilsInterop
+
+using ..DiagrammaticEquations: AbstractDecapode, Quantity
```
jpfairbanks, line 3: Where did Quantity come from?

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+import Base: +, -, *
+import Catlab: Δ, ∧
+
+# ##########################
+# ThDEC
+#
+# Type necessary for symbolic utils
+# ##########################
+
+abstract type ThDEC <: Quantity end
```
jpfairbanks, line 19: I'd rather the module be `ThDEC` and the type be `DECType` or `DECQuantity`

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+const DualVF{s,n} = VField{true,s,n}
+export DualVF
+
+# ACTIVE PATTERNS
+
+@active ActForm(T) begin
+    if T <: Form
+        Some(T)
+    end
+end
```
jpfairbanks, line 84: because ACT already means applied category theory. Could we use a one letter name for these like PForm(T)? P for pattern?

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+@active ActVFParams(T) begin
+    if T <: VField
+        Some([T.parameters...])
+    end
+end
+
+# HERE WE DEFINE THE SYMBOLICUTILS
+
+# for every unary operator in our theory, take a BasicSymbolic type, convert its type parameter to a Sort in our theory, and return a term
+unops = [:♯, :♭]
```
jpfairbanks, line 113: unused unops

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+end
+
+@alias (⋆₀, ⋆₁, ⋆₂, ⋆₀⁻¹, ⋆₁⁻¹, ⋆₂⁻¹) => ⋆
+
+@register Δ(S)::ThDEC begin
+    @match S begin
+        ActForm(x) => ⋆(d(⋆(d(x))))
+        _ => throw(SortError("Cannot take the Laplacian of $S"))
+    end
+end
```
jpfairbanks, line 142: register should return the function Δ and a rule that replaces Δ with ⋆(d(⋆(d(x)))) for 0-forms and ⋆(d(⋆(d(x)))) + d(⋆(d(⋆(x)))) for 1 forms. Then we should use it like

```julia
Δ, r_expand_Δ = @register Δ(S)::ThDEC begin
    @match S begin
        ActForm(x) => ⋆(d(⋆(d(x))))
        _ => throw(SortError("Cannot take the Laplacian of $S"))
    end
end
```

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+        (Scalar, Scalar) => Scalar
+        (Scalar, ActFormParams([i,d,s,n])) || (ActFormParams([i,d,s,n]), Scalar) => Form{i,d,s,n}
+        _ => throw(SortError("Cannot multiple $S1 and $S2"))
+    end
+end
+
+@register ∧(S1, S2)::ThDEC begin
+    @match (S1, S2) begin
+        (ActFormParams([i1,d1,s1,n1]), ActFormParams([i2,d2,s2,n2])) => begin
+            (d1 == d2) && (s1 == s2) && (n1 == n2) || throw(SortError("Can only take a wedge product of two forms of the same duality on the same space"))
```
jpfairbanks, line 175: these errors can be subtypes of `SortError`. So that the error message can be put into the show method for that subtype of `SortError`

src/deca/ThDEC.jl
```diff
@@ -0,0 +1,229 @@
...
+            else
+                throw(SortError("Can only take a wedge product when the dimensions of the forms add to less than n, where n = $n1 is the dimension of the ambient space: tried to wedge product $i1 and $i2"))
+            end
+        end
+    end
+end
+
+struct SortError <: Exception
+    message::String
+end
```
jpfairbanks, line 187: by making more subtypes of Sort error, you won't need a message field. The message can get baked into the subtype's show method.

src/symbolictheoryutils.jl
```diff
@@ -0,0 +1,161 @@
...
+        Expr(:(::), vars...) => ph.(vars)
+        Expr(:curly, :Type, Expr(:<:, Expr(:curly, type, params...))) => (type, params)
+        s => s
+    end
+    ph(body)
+    quote
+        $foo(arg, s1::B1, s2::B1) where {S1,S2,B1<:BasicSymbolic{S1},B2<:BasicSymbolic{S2}}
+
+    end
+end
```
jpfairbanks, line 146: Is this trying to define a show method for Sorts?

src/symbolictheoryutils.jl
```diff
@@ -0,0 +1,161 @@
...
+
+
+Expr(:=,
+     Expr(:where
+          [Expr(:call
+                foo,
+                Expr(:(::), e...)),
+           params...]),
+     Expr(:block, body...))
```
jpfairbanks, line 161: spurious code

src/symbolictheoryutils.jl
```diff
@@ -0,0 +1,161 @@
...
+end
+```
+
+Δ(S1, S2) begin
+    @match (S1, S2)
+        
+
+will create an additional method for Δ for operating on BasicSymbolic 
+"""
+macro register(head, body)
```
jpfairbanks, line 46: We should rename this to `@operator`

src/symbolictheoryutils.jl
```diff
@@ -0,0 +1,161 @@
...
+              @nospecialize
+              function $f($(basicsym_args...)) where {$(constraints_expr...)}
+                  s = $f($(getindex.(generic_vars, 2)...))
+                  SymbolicUtils.Term{s}($f ,[$(getindex.(basicsym_bindings, 1)...)])
+              end
+              export $f
+          end))
+
+    return result
+end
```
jpfairbanks, line 100: I think we need this function to also emit a `promote_symtype(::typeof{$f}, args...) = f(symtype.(args))` to make rules work.

jpfairbanks (changes requested):
Mostly bikeshedding things. With an inline comment about adding a method for `promote_symtype` to the registration macro

src/symbolictheoryutils.jl
quffaro: This will be removed in the next commit. Pushed whatever I had immediately.

src/symbolictheoryutils.jl
quffaro: This is spurious code that was helping me see (hence the name) the AST

src/deca/ThDEC.jl
quffaro: Left those to remind myself that they're unimplemented. I'll remove them in next comment

src/symbolictheoryutils.jl
```diff
@@ -1,15 +1,17 @@
 using MLStyle
 using SymbolicUtils
-using SymbolicUtils: Symbolic, BasicSymbolic, FnType, Sym
+using SymbolicUtils: Symbolic, BasicSymbolic, FnType, Sym, symtype
 
+""" ThDEC in DiagrammaticEquations must be subtyped by Number to integrate with SymbolicUtils. An intermediary type, Quantity, makes it clearer that terms in the theory are "symbolic quantities" which behave like numbers
```
jpfairbanks, line 5: `DECQuantities` in DiagrammaticEquations must be subtypes of Number to integrate with SymbolicUtils. An intermediary type, Quantity, makes it clearer that terms in the theory are "symbolic quantities" which behave like numbers. In the context of SymbolicUtils, a Number is any type that you can do arithmetic operations on.

src/SymbolicUtilsInterop.jl
quffaro: This is coming from `symbolictheoryutils`, which is not a module. Should I promote `symbolictheoryutils` to a module so it's clear where it's being imported from? Or should I even use `Quantity` at all? I like it as a clothesline to hang BasicSymbolic-integrated theories on, since they are treated as abstract/symbolic quantities which behave like numbers.

src/symbolictheoryutils.jl
```diff
@@ -1,15 +1,17 @@
...
 abstract type Quantity <: Number end
 export Quantity
 
 """
-Registers a new function
+Creates an operator `foo` with arguments which are types in a given Theory. This entails creating (1) a function which performs type construction and (2) a function which consumes BasicSymbolic variables and returns Terms.
 
 ```
-@register foo(S1, S2, ...)::ThDEC begin
+@operator foo(S1, S2, ...)::Theory begin
```
jpfairbanks, line 14: ahh, I see that this return type is like a categories as contexts approach. We are using the abstract type that all of our symbolic quantities inhabit as the "category as context" to anchor the algebraic theory on. I like this as a hybrid between the Catlab/Gatlab approach to explicit theories as the starting point and the SymbolicUtils "a number is a thing you can do math on" approach that minimizes the role of the algebraic theory.

src/symbolictheoryutils.jl
```diff
@@ -81,24 +83,33 @@ macro register(head, body)
...
-          end))
+        end))
+
+    push!(result.args,
+          esc(quote
+            # we want to feed symtype the generics
+            function SymbolicUtils.promote_symtype(::typeof($f), 
+                    $(bs_arg_exprs...)) where {$(constraint_exprs...)}
+                $f($(getindex.(generic_vars, 2)...))
+            end
```
jpfairbanks, line 107: i'm still worried about the overly liberal usage of `esc` here. Not confident that we don't need it here, but we should try to reduce the calls to esc when they aren't needed.

test/decasymbolic.jl
```diff
@@ -1,8 +1,8 @@
 using Test
-using DiagrammaticEquations.Deca.TheoryDEC
+using DiagrammaticEquations.Deca.ThDEC
 using DiagrammaticEquations.decapodes
 using SymbolicUtils
-using SymbolicUtils: symtype
+using SymbolicUtils: symtype, promote_symtype
```
jpfairbanks, line 5: do we need to import to extend?

jpfairbanks (changes requested):
Just the docstring and the esc usage comment blocking this

test/decasymbolic.jl
```diff
@@ -1,46 +1,78 @@
...
+    # TODO rewriting not working atm
+    # del_expand = Chain(del_expand0, del_expand1)
+
+    @test_throws Exception ∇(b)
+    @test symtype(∇(u)) == PrimalForm{0, :X ,2}
+
+    @test_broken promote_symtype(Δ, [u,v])
+
+    @test del_expand_0(u) == ★(d(★(d(u))))
 end
```
jpfairbanks, line 76: add a test for alias

jpfairbanks (changes requested):

Issue #74, opened 2024-09-12
@decapode not usable in source


lukem12345:
Attempting to define a particular Decapode in any source file (of the DiagrammaticEquations package) causes a segmentation fault.

To reproduce:
1. Go to any source file, such as `src/language.jl`
2. Define a decapode. e.g.
```julia
g = @decapode begin
  A == B + C
end
```
3. Open a REPL. Do `using DiagrammaticEquations, DiagrammaticEquations.Deca`. Attempt to construct a Decapode using the `@decapode` macro. This causes a segmentation fault.

lukem12345:
@jpfairbanks Is this an intertypes thing?

lukem12345:
Following the above steps on commit 88b18743976c55c3d77316e40af3117aec368f1b , no segfault occurs and everything runs as expected.

lukem12345:
Segfaults on 9bbe494bbb61f6d180c465ea8c834f6ce15a983f
Does not segfault on e4df0b6009696711733a0d33bdb5706b53dd6ba6

lukem12345:
@quffaro Can you take a gander at this and see if this is an issue for ACSets.jl?

quffaro:
@lukem12345 I was able to replicate the issue. Thank you for the helpful analysis! I'll look at the two commits you posted

jpfairbanks:
I bet this has to do with the fact that we don't put the Decapodes ACSet types into their own submodule. So there is no module boundary between code that is in a src/x.jl file and the code that defines the types. 

The tests and docs both live in modules that import DiagrammaticEquations rather than directly live inside it. So we can create decapodes outside that module, but not within it.

Hopefully the only fix we would need is to wrap any intertypes generated type with a submodule of the package.

lukem12345:
After reconsidering, #88 won't fix this, because it just changes the parser, but still uses the Decapode ACSet.

A simple way of resolving this issue is to move intertypes related code into the tests, where we can just verify that the DecapodeACSet from src is compatible.

jpfairbanks:
I think we could also add an extra module that wraps the intertype's and then re-export that functionality

lukem12345:
@jpfairbanks Tried this simple nesting structure but it failed. What nesting pattern do you mean?

https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/compare/main...refs/heads/llm/wrap_it

[...]

PR #75, opened 2024-09-13
Provide Collection of Rewrite Rules

lukem12345:
See https://github.com/AlgebraicJulia/Decapodes.jl/issues/252

This branch may be affected by #74

### DIFF
--- a/src/deca/deca_acset.jl
+++ b/src/deca/deca_acset.jl
@@ -197,15 +197,6 @@ function vec_to_dec!(d::SummationDecapode)
   d
 end
 
-# TODO: When SummationDecapodes are annotated with the degree of their space,
-# use dispatch to choose the correct set of rules.
-"""    function resolve_overloads!(d::SummationDecapode)
-
-Resolve function overloads based on types of src and tgt.
-"""
-resolve_overloads!(d::SummationDecapode) =
-  resolve_overloads!(d, op1_res_rules_2D, op2_res_rules_2D)
-
 # Default Rewrite Rules
 # ---------------------
 
@@ -213,7 +204,7 @@ rewrite_rules_2D = Vector{AbstractSDRewriteRule}([
   Op1SDRule(
     :Δ₀,
     @decapode begin
-      y == ∘(d,δ)(X)
+      y == ∘(d,⋆,d,⋆)(X)
     end),
 
   Op1SDRule(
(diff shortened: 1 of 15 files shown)


src/deca/deca_acset.jl
```diff
@@ -359,3 +359,71 @@ Resolve function overloads based on types of src and tgt.
...
+    @decapode begin
+      y == ∘(d,δ)(X)
+    end),
+
+  Op1SDRule(
+    :Δ₁,
+    @decapode begin
+      (X,y)::Form1
+      y == ∘(d,δ)(X) + ∘(δ,d)(X)
+    end),
```
jpfairbanks, line 377: Seems like overkill to use a whole decapode for this since it will always be a single LHS, RHS pair. Also the SymbolicUtils stuff will be able to use the `@rule` macro for defining rules.

lukem12345: This is to replace operations which take place directly on Decapodes, such as https://github.com/AlgebraicJulia/Decapodes.jl/blob/main/src/operators.jl#L320

lukem12345:
@copilot Rebase, then ensure tests pass.

src/deca/deca_acset.jl
lukem12345: Another advantage of this method is that the declarative rewrite system runs through the same `type_check` collection of type inference rules. Integrating into Decapodes, found a test that was ill-typed:

https://github.com/AlgebraicJulia/Decapodes.jl/blob/main/test/operators.jl#L23-L29

This was hard to uncover originally because the imperative version is:
```julia
function add_Inter_Prod!(d::SummationDecapode, proj1_Inter::Int, proj2_Inter::Int, res_Inter::Int)
  ## Adds the hodge Dual to Primal
  inv_hodge_tgt = add_part!(d, :Var, type=:infer, name=nothing)
  add_part!(d, :Op1, src=proj1_Inter, tgt=inv_hodge_tgt, op1=:⋆)

  ## Adds the wedge between Primal and Primal
  wedge_res = add_part!(d, :Var, type=:infer, name=nothing)
  add_part!(d, :Op2, proj1=inv_hodge_tgt, proj2=proj2_Inter, res=wedge_res, op2=:∧)

  ## Adds the hodge Primal to Dual
  add_part!(d, :Op1, src=wedge_res, tgt=res_Inter, op1=:⋆)
end

function add_Inter_Prod_1D!(::Val{1}, d::SummationDecapode, proj1_Inter::Int, proj2_Inter::Int, res_Inter::Int)
  add_Inter_Prod!(d, proj1_Inter, proj2_Inter, res_Inter)
end
```

PR #80, opened 2024-10-11
Fix show method for DEC symbolics

jpfairbanks:
@GeorgeR227, notes from today's meeting


GeorgeR227:
I tried to add a type to the `args` argument in our `show_call` to avoid piracy but I can't since the `sorted_arguments` used in the `show_term` returns as a `Vector{Any}`. In that case I'll probably just take the test code as given and just put it in source.

lukem12345:
This was fixed by various upstreamings

PR #83, opened 2024-10-17
Added steadystates dictionary keyword to `collate`

quffaro:
This feature adds a keyword to the `collate` function to accept a dictionary of symbols which specify restriction maps onto boundary conditions.


src/collages.jl
```diff
@@ -49,12 +47,16 @@ function collate(equations, boundaries, uwd, symbols)
     newtype = @match (gettype(equations, en), gettype(boundaries, bn)) begin
         (:infer, _) => :infer
```
lukem12345, line 48: I think this should map to the inferred type

test/collages.jl
```diff
@@ -141,3 +141,47 @@ TwiceCollage = DiagrammaticEquations.collate(
   name = [:K, :J, :K̇, :r1_J, :Jb1, :r2_J, :Jb2]
 end
 
+# Test that `steadystates` .
```
lukem12345, line 144: What is this test supposed to demonstrate in particular

src/collages.jl
```diff
@@ -49,12 +47,16 @@ function collate(equations, boundaries, uwd, symbols)
...
         (_, :infer) => :infer
-        (:Form0, :Constant) => :Form0
+        (_, :Constant) => :Constant
         (x, y) && if x == y end => x
         (x, y) => error("Type mismatch between $x and $y")
     end
     s_var = add_part!(f, :Var, type=newtype, name=bn)
     add_part!(f, :Op2, proj1=var, proj2=s_var, res=b_var, op2=uwd[b, :name])
+
+    if bn ∈ values(steadystates)
```
lukem12345, line 57: Are the keys of the dict ever used?

src/collages.jl
```diff
@@ -47,14 +47,17 @@ function collate(equations, boundaries, uwd, symbols)
     # Insert the "masking" operation.
     gettype(xs, n) = xs[only(incident(xs, n, :name)), :type]
     newtype = @match (gettype(equations, en), gettype(boundaries, bn)) begin
-        (:infer, _) => :infer
-        (_, :infer) => :infer
-        (:Form0, :Constant) => :Form0
+        (:infer, x) || (x, :infer) => x
+        (x, :Constant) || (:Constant, x) => x
```
lukem12345, line 51: Need to check for `Parameter` here too

quffaro: thank you, addressed in commit e32d5f7

Issue #81, opened 2024-10-11
Remove user annotations of DEC operators


GeorgeR227:
With the merging of the SymbolicUtils interoperability branch into DiagrammaticEquations we want to remove the user-allowed typing of DEC operators in the Decapode. For example, a user using `d₁` instead of the generic `d`. While this typing is still critical during the code generation phase, in order to differentiate which matrix implementations of `d` to use for example, for the purposes of type inference/rewriting it is increasingly adding to the overall complexity of the code without providing much benefit.

To remedy this, we should instead only work with generic operators on the user-side, so things like `d₁` are no longer viewed as a valid DEC operator and only `d` is. Then, for code generation only, we take the operator's name and it's surrounding type information to generate its type-specific name there for implementation differentiation.

lukem12345:
> ... without providing much benefit.

To clarify, annotations enforce the selection of particular versions of operators. Some DEC equations are ambiguous (there may exist multiple valid typings) in terms of which operators to use, and different versions of operators will exhibit different (non-trivial) numerical properties.

GeorgeR227:
Yes this is true. The intention here is to only expose the operator names that are "as generic as possible" to the user. Like in the case for `d`, given the type information, the `d` can be resolved to one and only one specific operator. So internally, it makes some sense to remove other name annotations / resolve to a canon name to simplify internal code since we're not losing information in this case. I guess what I'm getting at is this is stronger canonicalization, where not just bespoke aliases but typing info can be used to resolve to a canon name.

Currently we do not do any type inference / type checking based on specific operator types, although PR #84 could definitely fit that in scope. This is for ACSets though and on the Symbolic side we actually type even specific types the same as the general, so `d_0` given a `:Form1` and returning `:Form2` passes but erroneously so. I agree that this kind of type checking ability should be added although on what side (ACSet/Symbolic) I'm not sure.

Issue #53, opened 2024-06-25
Use structs for inference and overloading rules


GeorgeR227:
Currently we use named tuples to hold our rules for type inference and overloading but that means we have massive type signatures in each of our functions. We should instead just use structs to hold this same information.

Instead of something like `op2_rules::Vector{NamedTuple{(:proj1_type, :proj2_type, :res_type, :resolved_name, :op), NTuple{5, Symbol}}}` we could just have `op2_rules::Vector{Op2Rule}`, holding the same information.

We might even be able to tie groups of rules together, like type rules, with an abstract type that allows us to dispatch and choose the correct type inference application function.

jpfairbanks:
Yes that makes sense to me. If you have multiple functions that take namedtuples with the same signature, then make it a struct. It should be exactly the same performance-wise.

jpfairbanks:
I wrote this on my fork of SymbolicUtils.

```julia

abstract type AbstractRule end
abstract type InferenceRule end

struct InferenceRule1 <: InferenceRule
  src_type
  tgt_type
  op_names
end
InferenceRule1(t) = InferenceRule1(t...)

struct InferenceRule2 <: InferenceRule
  proj1_type
  proj2_type
  res_type
  op_names
end
InferenceRule2(t) = InferenceRule2(t...)

struct ResolutionRule <: AbstractRule
  src_type
  tgt_type
  resolved_name
  op
end
ResolutionRule(t) = ResolutionRule(t...)
```
which lets you do:
```julia
"""
These are the default rules used to do type inference in the 1D exterior calculus.
"""
op1_inf_rules_1D = InferenceRule1.([
  # Rules for ∂ₜ 
  (:Form0, :Form0, [:∂ₜ,:dt]),
  (:Form1, :Form1, [:∂ₜ,:dt]),

  # Rules for d
  (:Form0, :Form1, [:d, :d₀]),
  (:DualForm0, :DualForm1, [:d, :dual_d₀, :d̃₀]),

  # Rules for ⋆
  (:Form0, :DualForm1, [:⋆, :⋆₀, :star]),
  (:Form1, :DualForm0, [:⋆, :⋆₁, :star]),
  (:DualForm1, :Form0, [:⋆, :⋆₀⁻¹, :star_inv]),
  (:DualForm0, :Form1, [:⋆, :⋆₁⁻¹, :star_inv]),

  # Rules for Δ
[...]

PR #84, opened 2024-10-21
Introducing Operator types

GeorgeR227:
This PR is driven in part by two different desires. There is a past one, in issue #53, to generalize the inference/overloading rules since there was a lot of repetition. Then there is also the current one, in issue #81, to remove the users from adding in DEC operator annotations. 

My idea is to begin the process of having DEC operators as simply being Symbols in the ACSet to being an Operator node which is richer in that it has typing and naming information attached to it. So for example we'll no longer have `:d` floating around but `:d` that acts on `:Form0` and returns `:Form1` and has other aliases `:d_0, etc...`. This PR won't change the actual ACSet to use this since this will be massively breaking but it will be used heavily for our compilation code.

This data organization moves our ACSet code closer to that of a theory, where operators are not just names but are expected to act on and return certain types. This structure also allows us to present generic names to the user while having functions that can take in the struct information and output a unique name for the operator.

There is some future-proofing included which expects the similar shift from Symbol types, like `:Form0`, to our new Symbolic types. A goal I'd want to tackle but will leave out of scope for when our Symbolics code matures more is to merge these Operators with their Symbolic function representations.


src/acset.jl
```diff
@@ -554,38 +620,55 @@ function infer_types!(d::SummationDecapode, op1_rules::Vector{NamedTuple{(:src_t
...
-  for op2_idx in parts(d, :Op2)
-    proj1 = d[:proj1][op2_idx]; proj2 = d[:proj2][op2_idx]; res = d[:res][op2_idx]; op2 = d[:op2][op2_idx]
-    proj1_type = d[:type][proj1]; proj2_type = d[:type][proj2]; res_type = d[:type][res]
-    for rule in op2_rules
-      if op2 == rule[:op] && proj1_type == rule[:proj1_type] && proj2_type == rule[:proj2_type] && res_type == rule[:res_type]
-        d[op2_idx, :op2] = rule[:resolved_name]
+  d
+end
+
+function type_check(d::SummationDecapode, type_rules::AbstractVector{Operator{Symbol}})
```
lukem12345, line 639: The control flow of this function is hard to parse. If tree probably is called for here.

lukem12345: Actually, the control flow should be reworked such that all type errors are returned.

src/acset.jl
```diff
@@ -440,6 +442,50 @@ function filterfor_ec_types(types::AbstractVector{Symbol})
...
+  map(1:length(types)) do i
+    aliases = isempty(sp_aliases) ? g_aliases : vcat(g_aliases, sp_aliases[i])
+    Operator{Symbol}(types[i], repeat([types[i]], arity), op_name, aliases)
+  end
+end
+
+function arthimetic_operators(op_name::Symbol, broadcasted::Bool, arity::Int = 2)
+  @match (broadcasted, arity) begin
+    (true, 2) => bin_broad_arith_ops(op_name)
+    _ => error("This type of arthimetic operator is not yet supported or may not be valid.")
```
lukem12345, line 475: Little typo here in arithmetic

src/acset.jl
```diff
@@ -466,36 +512,99 @@ function infer_sum_types!(d::SummationDecapode, Σ_idx::Int)
...
 
-  check_op = (d[op1_id, :op1] in rule.op_names)
-  if(check_op && (score_src + score_tgt == 1))
-    mod_src = safe_modifytype!(d, d[op1_id, :src], rule.src_type)
-    mod_tgt = safe_modifytype!(d, d[op1_id, :tgt], rule.tgt_type)
-    return mod_src || mod_tgt
+  if name_present && type_diff == 1
+    vars = vcat(edge_inputs(d, op_id, edge_val), edge_output(d, op_id, edge_val))
+    types = vcat(rule.src_types, rule.res_type)
+    return any(map(zip(vars, types)) do (var, type)
```
lukem12345, line 549: Is the zip necessary here?

GeorgeR227: No but it seems to only work without the zip if I remove the parantheses on the vars after the `do`.

lukem12345: Ok, this can get changed on the next commit

GeorgeR227: I actually already changed it in https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/84/commits/b4ba78701f36e31791f862dffc3e0f5d216fa296.

src/acset.jl
```diff
@@ -466,36 +512,105 @@ function infer_sum_types!(d::SummationDecapode, Σ_idx::Int)
   return applied
 end
 
-function apply_inference_rule_op1!(d::SummationDecapode, op1_id, rule)
-  score_src = (rule.src_type == d[d[op1_id, :src], :type])
-  score_tgt = (rule.tgt_type == d[d[op1_id, :tgt], :type])
+"""
+    check_operator(d::SummationDecapode, op_id, rule, edge_val; check_name::Bool = false, check_aliases::Bool = false, ignore_infers::Bool = false, ignore_usertypes::Bool = false)
```
jpfairbanks, line 516: docstring formatting

test/language.jl
```diff
@@ -856,13 +842,37 @@ end
...
+    C::Form0
+
+    B == fake(A)
+    D == fake2(A, C)
   end
+  op_not_in_rules_res = deepcopy(op_not_in_rules)
+  infer_types!(op_not_in_rules)
+
+  @test op_not_in_rules == op_not_in_rules_res
+  @test op_not_in_rules !== op_not_in_rules_res
```
jpfairbanks, line 859: how can both of these tests pass?

test/language.jl
```diff
@@ -1034,83 +1155,12 @@ end
...
-      summand = [3, 17]
-      summation = [1, 1]
-      sum = [5]
-      op1 = [:∂ₜ, :d₀, :d₀, :♯ᵖᵖ, :mag, :⋆₁, :dual_d₁, :⋆₀⁻¹]
-      op2 = [:*, :-, :^, :∧₁₀, :^, :∧₁₀]
-      type = [:Form0, :Form1, :Constant, :Form0, :infer, :Form1, :Form1, :Form1, :Form1, :Form0, :Form0, :PVF, :Form1, :infer, :Literal, :Form0, :Literal, :DualForm1, :DualForm2]
-      name = [:h, :Γ, :n, :ḣ, :sum_1, Symbol("•2"), Symbol("•3"), Symbol("•4"), Symbol("•5"), Symbol("•6"), Symbol("•7"), Symbol("•8"), Symbol("•9"), Symbol("•10"), Symbol("1"), Symbol("•11"), Symbol("2"), Symbol("•_6_1"), Symbol("•_6_2")]
-    end
-  end
-
```
jpfairbanks, line 1113: Do we not need to test this anymore?

GeorgeR227: It was figured out that this Decapode was actually poorly typed and thus not a good test for ensuring the overloading/inferring was working correctly. Instead I've converted this into a type-checking test. It's been shifted around I but not deleted.

test/language.jl
GeorgeR227: Line 858 is checking that the ACSets have the same entries, Line 859 is checking that the memory locations are not the same.

src/acset.jl
```diff
@@ -440,6 +442,50 @@ function filterfor_ec_types(types::AbstractVector{Symbol})
...
+
+function bin_broad_arith_ops(op_name)
+  all_ops = map(t -> Operator{Symbol}(t, [t, t], op_name), FORM_TYPES)
+  for type in vcat(USER_TYPES, NUMBER_TYPES)
+    append!(all_ops, map(t -> Operator{Symbol}(t, [t, type], op_name), FORM_TYPES))
+    append!(all_ops, map(t -> Operator{Symbol}(t, [type, t], op_name), FORM_TYPES))
+  end
+
+  all_ops
+end
```
lukem12345, line 484: This looks like it is trying to emulate the style of having an abstract type for an "Operator", can this idiom be used explicitly.

lukem12345:
@quffaro This PR checks types via a notion of a “score” which [counts](https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/84/files#diff-193c1fdaed9c36d8a8ce0b7e35147ca00bddc1b93ae40c511596a82a4675730cR527) the number of mis-matched types between a function’s type signature vs. a collection of type signatures (“rules”). It seems to be converging on [some variation](https://github.com/AlgebraicJulia/DiagrammaticEquations.jl/pull/84/files#diff-193c1fdaed9c36d8a8ce0b7e35147ca00bddc1b93ae40c511596a82a4675730cR605) of the “pattern matching” idiom. (If a pattern is fully matched, or no partial match is found (i.e. fall through), throw an error. Else when a partial match is found, throw an error.)

And so I am concerned that we may be unwittingly getting into maintaining a bespoke type checker by way of reimplementing of a pattern matching system. I think that the explicit notion of “scoring” is a step in the right direction, but ultimately should be realized by making use of a fully-featured pattern matching capabilities offered by MLStyle or a similar package like Moshi.

Can I get your thoughts on this? @quffaro

src/acset.jl
```diff
@@ -440,6 +439,88 @@ function filterfor_ec_types(types::AbstractVector{Symbol})
...
+        types1 = vcat(rule1.res_type, rule1.src_types)
+        types2 = vcat(rule2.res_type, rule2.src_types)
+
+        if length(types1) != length(types2)
+          continue
+        end
+
+        score = mapreduce(+, types1, types2; init = 0) do type1, type2
+          if type1 == type2
+            return 0
```
lukem12345, line 506: This function should be unpacked. It can return an Int, a Float, or a Bool

src/acset.jl
```diff
@@ -440,6 +439,88 @@ function filterfor_ec_types(types::AbstractVector{Symbol})
...
+    append!(all_ops, map(t -> Operator{Symbol}(t, [t, type], op_name), FORM_TYPES))
+    append!(all_ops, map(t -> Operator{Symbol}(t, [type, t], op_name), FORM_TYPES))
+  end
+
+  all_ops
+end
+
+# TODO: This could probably be implemented using a better version of `check_operator`
+# TODO: Add printing of rules which are ambigious with each other
+function check_rule_ambiguity(type_rules::AbstractVector{Operator{Symbol}})
```
lukem12345, line 488: This function is a little difficult to parse. Can this be broken down into component functions?

GeorgeR227:
I've rebased this PR and tests seem to be running. However, I've had to remove this rule, 
```Operator(:DVF, :Form1, :♯ᵖᵈ, [:♯])```
as including it caused some ambiguity tests to fail, since this rule is also included,
```Operator(:PVF, :Form1, :♯ᵖᵖ, [:♯])```.

The main reason is the alias along with the input type. For example, if in the Decapode, I'm given an input with `:Form1` and the operator is ```♯```, I'd have no idea if I should choose the output to be `:DVF` or `:PVF`. I think there are similar problems with the wedge that we get around by not having these generic aliases.

Luckily, this failing shows that the functionality of the rule ambiguity checking is working exactly as intended!

lukem12345:
Yes the notion of "ambiguity" as implemented in code here is a little too strict, since there may be multiple valid discretizations for an operator.

GeorgeR227:
Yeah, I think you've hit on the solution which really lies in overhauling the type inference/operator overloading.

Like you said, we could make the order we apply rules in more concrete, like by assigning some priority. Thus ambiguity could also check similar rules don't have the same priorities. Then an issue with this is that we'd have to make sure there's overall consistency. For example, if in a Decapode we have `:Form1` -> `♯` -> `infer` -> `♭ᵈᵖ` -> `infer`, then we might make a mistake if we are inferring through the `♯` and we assigned higher priority to `:♯ᵖᵖ` instead of `:♯ᵖᵈ`. 

I think in this case we could run only rules that are completely unambiguous and if a rule is ambiguous, we still include those in the running but don't apply it, with the hope that there might be enough information later to disambiguate it. So essentially we no longer run a greedy rule application but check all rules that may apply and play it safe by only applying if there is a single result. So in the above, if we later run the rule for `♭ᵈᵖ` and infer all the types above, then there is no more ambiguity. Then if, when we do type inference/operator overloading, the only rules left to apply are ambiguous or don't apply at all, we can stop there.

I think also valid for the flat and sharp is to make sharp_flat operator, `♭♯`, a supported operator and in this case choose the underlying implementation that works the best.

lukem12345:
Yes a two-staged process, each greedy, sounds best. The first stage of rules that must be applied, and the second of those ambiguous rules which agree in degree but not primality. This two step process can be repeated until no further rules are applied. Luckily the existing infrastructure is already centered around applying sets of rules, so this is purely a higher-level scheduling abstraction.

PR #89, opened 2025-02-04
WIP: PEG / String Macro for Decapodes

### LINKED ISSUE #88: PEG / String macro for Decapodes
Taken from Dylan and Christian’s design doc:

> [The Decapodes DSL is] built using a syntactic macro and a hand-written parser. The core problem with this technique lies in three areas.

> Because Julia’s compiler has the unicode star symbol [for example] defined as an operator… we are confined to the set precedence and properties that Julia assumes the star operator to have.

> The second problem lies in maintainability. Because we build a hand-written parser for each
DSL, developers have to deal with the complexity and bugs that arise from a complicated
program such as a parser. A small design change could cause hours of head-aches as one
attempts to modify a heavily recursive program, dealing with the mundane bugs that arise from
such.

> The third problem lies in extendability. Because each DSL requires a hand-written parser, we are
unable to reuse core components to quickly and easily create new DSLs.

> The solution is as follows. Our team will develop parsing expression grammars that will replace
the hand-written parsers and utilize string macros over syntactic macros. A parsing expression
grammar or a PEG is a formal grammar used for recognizing strings in the language defined
(“Parsing expression grammar”, 2024). PEGs are defined by a collection of parsing expressions
that can reference each other. Through this, we can rewrite an existing parser as a PEG where
each recursive parsing function is a parsing expression.

To be closed by PR associated with https://github.com/cscaff/DiagrammaticEquations.jl/tree/PEG-DSL

cscaff:
resolves #88 

Work in progress parsing expression grammar and string macro for Decapodes. The incentives are noted in the linked issue.


src/parser.jl
```diff
@@ -0,0 +1,276 @@
...
+@rule Ident = r"""([^0-9\+\*:{}→\n;=,\-−¦⊕⊖⊞⊟∪∨⊔±∓∔∸≏⊎⊻⊽⋎⋓⟇⧺⧻⨈⨢⨣⨤⨥⨦⨧⨨⨩⨪⨫⨬⨭⨮⨹⨺⩁⩂⩅⩊⩌⩏⩐⩒
+  ⩔⩖⩗⩛⩝⩡⩢⩣\\\/⌿÷%&··⋅∘×∩∧⊗⊘⊙⊚⊛⊠⊡⊓∗∙∤⅋≀⊼⋄⋆⋇⋉⋊⋋⋌⋏⋒⟑⦸⦼⦾⦿⧶⧷⨇⨰⨱⨲⨳⨴⨵⨶⨷⨸⨻⨼⨽⩀<⩃⩄⩋⩍⩎⩑⩓⩕⩘⩚⩜
+  ⩞⩟⩠⫛⊍▷⨝⟕⟖⟗⨟\^↑↓⇵⟰⟱⤈⤉⤊⤋⤒⤓⥉⥌⥍⥏⥑⥔⥕⥘⥙⥜⥝⥠⥡⥣⥥⥮⥯￪￬\|\(\)\s][^\+\*:{}→\n;=,\-−¦⊕⊖⊞⊟∪∨⊔±∓∔∸≏⊎⊻⊽⋎⋓⟇⧺⧻
+  ⨈⨢⨣⨤⨥⨦⨧⨨⨩⨪⨫⨬⨭⨮⨹⨺⩁⩂⩅⩊⩌⩏⩐⩒⩔⩖⩗⩛⩝⩡⩢⩣\\\/⌿÷%&··⋅∘×∩∧⊗⊘⊙⊚⊛⊠⊡⊓∗∙∤⅋≀⊼⋄⋆⋇⋉⋊⋋⋌⋏⋒⟑⦸⦼⦾⦿⧶⧷⨇⨰
+  ⨱⨲⨳⨴⨵⨶⨷⨸⨻⨼⨽⩀<⩃⩄⩋⩍⩎⩑⩓⩕⩘⩚⩜⩞⩟⩠⫛⊍▷⨝⟕⟖⟗⨟\^↑↓⇵⟰⟱⤈⤉⤊⤋⤒⤓⥉⥌⥍⥏⥑⥔⥕⥘⥙⥜⥝⥠⥡⥣⥥⥮⥯￪￬\|\(\)\s]*)""" |> v -> Symbol(v)
+
+# Digits consist of numerical characters
+@rule Digit = r"([\-]?)([0-9]+)(\.[0-9]+(e[0-9]+)?)?" |> v -> Lit(Symbol(v))
+
+# Unary operators support operators that work on one operand525 170 8618
```
lukem12345, line 124: Those numbers should be deleted.

jpfairbanks:
@lukem12345 is this ready to merge?

lukem12345:
> @lukem12345 is this ready to merge?

I cleaned the test file and fixed the Project.toml setup. I'm going to look at the source file again. We shouldn't merge this into main until Catlab triggers a release.

src/parser.jl
```diff
@@ -0,0 +1,276 @@
...
+  ⩔⩖⩗⩛⩝⩡⩢⩣\\\/⌿÷%&··⋅∘×∩∧⊗⊘⊙⊚⊛⊠⊡⊓∗∙∤⅋≀⊼⋄⋆⋇⋉⋊⋋⋌⋏⋒⟑⦸⦼⦾⦿⧶⧷⨇⨰⨱⨲⨳⨴⨵⨶⨷⨸⨻⨼⨽⩀<⩃⩄⩋⩍⩎⩑⩓⩕⩘⩚⩜
+  ⩞⩟⩠⫛⊍▷⨝⟕⟖⟗⨟\^↑↓⇵⟰⟱⤈⤉⤊⤋⤒⤓⥉⥌⥍⥏⥑⥔⥕⥘⥙⥜⥝⥠⥡⥣⥥⥮⥯￪￬\|\(\)\s][^\+\*:{}→\n;=,\-−¦⊕⊖⊞⊟∪∨⊔±∓∔∸≏⊎⊻⊽⋎⋓⟇⧺⧻
+  ⨈⨢⨣⨤⨥⨦⨧⨨⨩⨪⨫⨬⨭⨮⨹⨺⩁⩂⩅⩊⩌⩏⩐⩒⩔⩖⩗⩛⩝⩡⩢⩣\\\/⌿÷%&··⋅∘×∩∧⊗⊘⊙⊚⊛⊠⊡⊓∗∙∤⅋≀⊼⋄⋆⋇⋉⋊⋋⋌⋏⋒⟑⦸⦼⦾⦿⧶⧷⨇⨰
+  ⨱⨲⨳⨴⨵⨶⨷⨸⨻⨼⨽⩀<⩃⩄⩋⩍⩎⩑⩓⩕⩘⩚⩜⩞⩟⩠⫛⊍▷⨝⟕⟖⟗⨟\^↑↓⇵⟰⟱⤈⤉⤊⤋⤒⤓⥉⥌⥍⥏⥑⥔⥕⥘⥙⥜⥝⥠⥡⥣⥥⥮⥯￪￬\|\(\)\s]*)""" |> v -> Symbol(v)
+
+# Digits consist of numerical characters
+@rule Digit = r"([\-]?)([0-9]+)(\.[0-9]+(e[0-9]+)?)?" |> v -> Lit(Symbol(v))
+
+# Unary operators support operators that work on one operand
+@rule UnaryOperator = PrecMinusOp , PrecDivOp , PrecPowerOp
```
lukem12345, line 125: Of these, only `-` can be used as a unary operation. Which tests rely on these other rules being listed as unary operations?

src/parser.jl
```diff
@@ -0,0 +1,276 @@
...
+"""
+function BuildApp2(v)
+    # Creates array of operations
+    result = vcat(v[1], map(x -> [x[2], x[end]], v[2])...)
+    # Converts array to tree structure
+    while length(result) > 1
+      Applied = App2(Symbol(result[2]), result[1], result[3])
+      result = vcat([Applied], result[4:end])
+    end
+    return result[1]
```
lukem12345, line 244: This iteration scheme is difficult to understand. Currently, this algorithm is iterating over an array that is being modified in-place with `vcat`, and the check for termination also depends on the length of this `vcat`ed array. This can be improved by using a recursive algorithm, writing the output into a new variable or array, and/ or iterating over the initial array multiple elements at a time (e.g. grabbing elements 2-at-a-time).

lukem12345:
@cscaff Next time, you should rebase your local branch, add your commit, and then push just your single commit. Committing code that undoes prior commits and then merging makes it difficult to track changes and increases noise.

lukem12345:
It looks like a consequence of the current setup, related to the `Operation`, `SummationOperation`, and `MultOperation` rules, is that `+` and `*` cannot be used as prefix operations.
e.g.
```julia
decapode"
  A == *(B, C)
"
```
will throw a `ParseError`.

@cscaff Are you able to easily amend the current code to fix this, or should I open another issue for this behavior

lukem12345:
>> I've been trying to get "+" added but it conflicts with the current usage. I'm trying to understand it. A more brute alternative is also to just add "+" as an option in CallName since that is what makes the name in the prefix notation.

If the issue is just that it will be parsed in such a case as an `Op2`, that's ok. Otherwise, you can commit your fix for `*` alone here, and we can note the difference with `+` for a follow-up issue.

cscaff:
> > I've been trying to get "+" added but it conflicts with the current usage. I'm trying to understand it. A more brute alternative is also to just add "+" as an option in CallName since that is what makes the name in the prefix notation.
> 
> If the issue is just that it will be parsed in such a case as an `Op2`, that's ok. Otherwise, you can commit your fix for `*` alone here, and we can note the difference with `+` for a follow-up issue.

Ok! I'll do this today!

lukem12345:
@cscaff Can you add a test for the prefix notation case? A good practice when handling something like this is to add a test that demonstrates the behavior that is brought up. This way we can be sure that the code does what we expect and guard against [regressions](https://en.wikipedia.org/wiki/Regression_testing).

test/parser.jl
```diff
@@ -0,0 +1,747 @@
...
+    App1(:f, Plus([Var(Symbol("n")), Lit(Symbol("1"))]))
+  @test Call("f(n+1, N+2)")[1] ==
+    App2(:f,
+      Plus([Var(Symbol("n")), Lit(Symbol("1"))]),
+      Plus([Var(Symbol("N")), Lit(Symbol("2"))]))
+  @test Call("⊕(a, b)")[1] == App2(:⊕, Var(Symbol("a")), Var(Symbol("b")))
+  @test Call("⊕(a)")[1] == App1(:⊕, Var(Symbol("a")))
+  @test Call("HI(a, b)")[1] == App2(:HI, Var(Symbol("a")), Var(Symbol("b")))
+  @test Call("d(Ψ)")[1] == App1(:d, Var(Symbol("Ψ")))
+  @test Call("(d)(Ψ)")[1] == App1(:d, Var(Symbol("Ψ")))
```
lukem12345, line 198: This test is misleading with how such a string is parsed in the context of the string literal.
Here, It is appropriately parsed as an `App1` since it is a unary operator with parentheses around the operator.
However, when such a string appears in a `decapode""`, this is actually parsed as a multiplication:
```julia-repl
julia> parse_whole(DecapodeExpr, "A == (d)(ψ)\n")
DecaExpr(Judgement[], Equation[Eq(Var(:A), App2(:*, Var(:d), Var(:ψ)))])
```

cscaff: Huh, I thought an end-end accounted for that test. Call has higher precedence than multiplication so naturally that should parse into a call, but I'll look and see what's going on.

cscaff: This passes: ```  parsing_is_isomorphic("
    (HT, Tₛ)::Form0
    (D, cosϕᵖ, cosϕᵈ)::Constant
    
    HT == (D ./ cosϕᵖ) .* (⋆)(d(cosϕᵈ .* (⋆)(d(Tₛ))))")``` which is weird if that call feature does not work, but I do see where it could fail. Not sure what's going on just yet.

cscaff: Ok, I guess the above example was actually multiplication? It definitely was not working in situations of "(call)(arg)". I moved around some rules within multiplication and grouping terms to ensure it always parses "(call)(arg)" as an app1 before a mult. Tests are included.

lukem12345:
@cscaff I noticed that none of the end-to-end tests used the multi-line comment syntax. The following Decapode is not parsed:
```julia
decapode"
       #=
         Multi-line comment
       =#
       A == d(B)
       "
```

Note that the following one does parse correctly:
```julia
decapode"
       A == d(B)
       #=
         Multi-line comment
       =#
       "
```

lukem12345:
@cscaff I added a fix for the scientific floats issue. The issue with Decapodes that start with multi-line comments is still outstanding.

cscaff:
> @cscaff I added a fix for the scientific floats issue. The issue with Decapodes that start with multi-line comments is still outstanding.

Ok! Thanks! Sorry for the silence. I'm trying to finish up a macro for sheaves to include in my senior project presentation. Once I get senior project officially done, finals week will be pretty free for me to go through bugs and minor changes.

lukem12345:
This looks good to merge pending the release of the next version of Catlab.

PR #90, opened 2025-02-05
Coloring Subdecapodes

quffaro:
The `oapply` feature allows Decapodes to be composed together according to a predefined UWD. However the relationship between submodels and the model is not preserved in the cospan, as the legs are empty. Regardless we lose any formal reference to the submodels when we call `apex` on the resulting cospan. Instead, we track the namespace by prefixing unshared variable terms by the name of the box on the UWD they slot into. While its possible to promote our naming convention to a vector of symbols, we can still generally retrieve the origin of the decapode through the underscore convention. Sometimes, we can rely on the graph embedding GraphViz chooses as also a heuristic for separating out submodels. A good example of this is the Klausmeier model (below).

It can be helpful if not pleasant for submodels to be colored. This draft PR begins as a proof-of-concept that we can color submodels. Variables that are not shared (prefixed by their box) are colored.

**Some limitations:**  We do not have a way to specify the color of a box from the relation diagram, as this might require `set_vprop!` call to a box in the relation after it has been created, and in turn extend our scope to Catlab.jl. We do not yet have a way to have colors persist across `oapply`. The current implementation picks  (without repetition)  a few colors I deemed "pleasant," so if we combine a number of Decapodes greater than the color set, then we will encounter an error.

klausmeier

bsh

nhs


src/deca/deca_visualization.jl
```diff
@@ -51,9 +53,28 @@ varname(d, v, verbose) = begin
   return "$name:$(spacename(d, v))"
 end
 
+const pleasant_colors = [:honeydew2, :tomato1, :darkorange1, :darkgoldenrod1,
```
lukem12345, line 56: Instead of throwing an error if not enough colors are available, we should `@warn` that colors are repeating, or nodes will stop being colored. Alternatively, we can linearly sample some gradient.

See https://juliagraphics.github.io/ColorSchemeTools.jl/latest/makingschemes/

lukem12345 (changes requested):

lukem12345:
@quffaro Coloring should be controlled by a flag, like `verbose`.

quffaro:
klausmeier_blue

This commit randomly samples a pastel color scheme. Because GraphViz accepts only color words ("white,", "green", e.g.), it uses the Euclidean norm to find the color word whose color value is the closed to a given sampled color.

jpfairbanks:
You can color things with HEX codes in graphviz 

https://graphviz.org/docs/attr-types/color/

It takes the following most common web format of RGBA, by setting A to approximately .3, most colors become a nice background color that isn't too dark to read black text on.

> "#%2x%2x%2x%2x" | Red-Green-Blue-Alpha (RGBA)

quffaro:
jl_N8zq3sd65u
Converted to hex codes and stripped away code to approximate given pastels to color words.

lukem12345:
There are a few ways to improve the current code:

1. Demonstration of this feature in the DiagrammaticEquations docs.
2. Demonstration of this feature in Decapodes docs.
3. Handle coloring diagrams with more than 256 components.
4. Better checking of color codes that are too close.
5. Better checking of color codes that are too close to black.
6. Deterministic color. This would e.g. enable composition diagrams and Decapodes to share consistent colors. One means would be to simply sort namespaces alphabetically, and index into the ColorScheme with regular ranges instead of randomly.

As it is an opt-in feature, and the basic issues have been addressed in most cases, this code is fine to merge.

PR #94, opened 2025-03-01
Replacing `rvrv` code pattern

quffaro:
This PR addresses a TODO in `acsets.jl` for replacing an array of `incident` calls with an ACSet query.
```julia
chain_starts = (From(:Op1) |> 
  Where(:src, d[:res] ∪ d[:sum] ∪ filter(!isnothing, infer_states(d))) |
  Where(:src, From(:Op1=>:tgt) |> Where(:op1, collect(black_list))))(d)
```

This PR depends on [FunSQL-like querying an ACSet](https://github.com/AlgebraicJulia/ACSets.jl/pull/163).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,8 +1,8 @@
 name = "DiagrammaticEquations"
 uuid = "6f00c28b-6bed-4403-80fa-30e0dc12f317"
 license = "MIT"
-authors = ["James Fairbanks", "Andrew Baas", "Evan Patterson", "Luke Morris", "George Rauta", "Matt Cuffaro"]
 version = "0.2.3"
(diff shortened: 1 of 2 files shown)


lukem12345:
@quffaro The query is failing due to trying to read something at index 1 from an empty buffer.

lukem12345:
@quffaro Looks like we could use another match here:

https://github.com/AlgebraicJulia/ACSets.jl/blob/fc42abaf3ac1ee453ebf91aea3692694b6929ae4/src/Query.jl#L164

`(_, []::ACSetSQLNode)` or similar perhaps.

PR #99, opened 2025-04-21
Interpreting CCL JSON as ACSet

quffaro:
CatColab currently the ability to export its various datatypes as JSON, including the model of its partial implementation of the DEC as well as diagrams in it (Decapodes). With some kinks needing to be sorted out, this PR is a proof-of-concept that specific AlgebraicJulia libraries can interpret a "CCLACSet" into their relevant datatype.

Concretely, we intend to have a method of `SummationDecapode` which can ingest a `UUIDLabeledGraph` and return a Decapode.


src/deca/deca_acset.jl
```diff
@@ -359,3 +359,22 @@ Resolve function overloads based on types of src and tgt.
...
+    end
+    foreach(parts(uuidlg, :E)) do part
+        src, tgt, elabel = subpart.(Ref(uuidlg), Ref(part), [:src, :tgt, :elabel])
+        add_part!(d, :Op1, src=src, tgt=tgt, op1=elabel)
+        if elabel == "∂ₜ"
+            add_part!(d, :TVar, incl=tgt)
+        end
+    end
+    d
+end
```
lukem12345, line 379: Something like this would be nice:
```julia
function SummationDecapode(uuidlg::UUIDLabeledGraph)
  @acset SummationDecapode{Any, Any, Symbol} begin
    V = nparts(uuidlg, :V)
    vtype = map(x -> something(x, :infer), uuidlg[:vtype])
    vlabel = Symbol.(uuidlb[:vlabel])

    E = nparts(uuidlg, :E)
    src = uuidlg[:src]
    tgt = uuidlg[:tgt]
    elabel = uuidlb[:elabel]

    TVar = count(==("∂ₜ"), uuidlb[:elabel])
    incl = findall(==("∂ₜ"), uuidlb[:elabel])
  end
end
```

quffaro: Thanks @lukem12345, this is a great suggestion. I've implemented it.

PR #102, opened 2025-05-14
Update UWD import for Catlab 0.16.20. Force GATlab 0.1.5

quffaro:


Project.toml
```diff
@@ -20,7 +20,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 [compat]
 ACSets = "0.2 - 0.2.21, ~0.2.23"
 Catlab = "0.15, 0.16"
-Colors = "0.12.11, 0.13.0"
```
lukem12345, line 23: This should not be done here, but rather in GATlab, right?

Project.toml
```diff
@@ -20,7 +20,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
...
+Colors = "0.12.11"
```
lukem12345, line 23: This should not be done here, but rather in GATlab, right?

jpfairbanks: You are saying that we shouldn't specify a colors compat at all in this package? It is listed as an explicit dependency at the top of the Project.toml file, so I think it needs a compat version.

Project.toml
```diff
@@ -9,6 +9,7 @@ ACSets = "227ef7b5-1206-438b-ac65-934d6da304b8"
 Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
+GATlab = "f0ffcf3b-d13a-433e-917c-cc44ccf5ead2"
```
jpfairbanks, line 12: Do we need an explicit Gatlab import? If we get everything through Catlab, that wouldn't be necessary.

Project.toml
lukem12345: No, just the removal of the 0.13.0 version which DiagrammaticEquations is compatible with

Project.toml
lukem12345: That's right, this is also a workaround for forcing Catlab to use the most up-to-date version of Catlab

lukem12345:
I'm going to merge this PR because I do not want builds of DiagrammaticEquations to fail on main. But adding GATlab as an explicit dependency to manually set its compat should be removed at some point. This points to there being some unresolved compat that should be set upstream, possibly breaking buiilds for unknown other packages.

PR #110, opened 2025-11-17
Bump Catlab compat to 0.17

quffaro:


lukem12345 (approved):
Approve. I think Catlab should rename the parser function it calls `collect` in some future release.

PR #138, opened 2026-06-11
Support SymbolicUtils v4, dropping v3

lukem12345:
PR #109 allowed SymbolicUtils v4 in the Project toml, but v4 was not getting picked up by the tests due to compat constraints (apparently due to DataStructures v0.19.) After clearing these compat complaints upstream in various PRs, this unblocked SymbolicUtils v4, and associated tests here started to fail.

(This PR also serves as a test of the recent Claude Fable model release.)

Claude summary follows:
```
Support SymbolicUtils v4, dropping v3

SymbolicUtils v4 reworked BasicSymbolic: the abstract Symbolic type and
head are gone, the type parameter is now a vartype (SymReal) rather than
the symtype, and the symtype moved to a runtime field passed via the
`type` keyword. Constants now arrive wrapped in Const variants.

- Replace Symbolic with BasicSymbolic; head with operation.
- Construct Sym/Term as Sym{SymReal}(name; type=T) and
  Term{SymReal}(f, args; type=S).
- Unwrap Const literals in decapodes.Term, and compare operations with
  === since == on a symbolic operation now builds a symbolic equation.
- Stop emitting call methods for +, -, * from @operator: the unary
  minus was an exact overwrite of SymbolicUtils's own method, and v4
  (like v3) already routes result types through promote_symtype.
- Type real coefficients like literals (v4 lowers -x to -1 * x), and
  permit products of identically-typed forms, since v4's mul worker
  folds promote_symtype(*, S, S) when typing scalar multiples like 2x.
- Update two tests for v4's different canonical ordering of sums.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
```


lukem12345:
More Claude output:

```
# Symbolics interop review (SymbolicUtils v4)

Sweep over `symbolictheoryutils.jl`, `SymbolicUtilsInterop.jl`, `acset2symbolic.jl`,
and `deca/ThDEC.jl`. Each bug was reproduced before fixing.

## Fixed

1. **`rules(f)` one-arg convenience was broken** — `@operator` generated
   `rules(f, Val{1})`, passing the type `Val{1}` instead of the instance `Val(1)`
   (and hardcoded arity 1). `rules(Δ)` threw a `MethodError`. Now `Val($arity)`.

2. **`@alias` `nameof` methods were broken** — the generated method accepted only one
   symtype argument, so binary aliases like `∧₀₁` threw in `decapodes.Term`'s `App2`
   branch. Worse, the body `Symbol("$alias")` never interpolated at macro time and
   referenced an undefined runtime variable. Now
   `Base.nameof(::typeof($alias), s...) = $(QuoteNode(alias))`.

3. **`@alias` didn't forward `promote_symtype`** — `promote_symtype(d₀, u)` threw,
   breaking `to_symbolics` on decapodes carrying subscripted ops (e.g. after overload
   resolution). Aliases now forward `promote_symtype` to their rep.

4. **Pirated `ComposedFunction` method narrowed** — the old
   `promote_symtype(::ComposedFunction, args)` was type piracy (the TODO was right),
   and SymbolicUtils v4 ships an equivalent, more-specific method on symtypes that
   wins dispatch. Replaced with a narrow lift for symbolic arguments only
   (`promote_symtype(d ∘ d, u)` is tested API), delegating to SymbolicUtils.

5. **Literals in sums degraded the type** — `symtype(u + 1)` was `Number`, and the
   degradation propagated (`d(u + 1)::Number`). Real summands now promote like
   `Literal` for `+` (mirroring the existing `*` rules); binary `-` is covered since
   `a - b` lowers to `a + (-b)`. Now `symtype(u + 1) == Form0`.

6. **Dead branch removed** — `op === DerivOp` in `decapodes.Term` could never be true
   (`DerivOp` is a `Symbol`; `operation(t)` is always a function or symbolic).

## Known limitations (not addressed)

- `Lit` converts to a raw Julia number, so an equation side that is *only* a literal
  cannot construct a `SymbolicEquation{BasicSymbolic}`.
- Space/dimension are hardcoded (`space = :I`, `dim = 2`) in `decavar_to_symbolics`
  and `DecaExpr`.
- Docstring nits: `@operator` shows `::typeof{f}` and uses `⋆` for `★`; the
  `SymbolicUtils.BasicSymbolic` example passes type *instances* where a
  `Dict{Symbol,DataType}` is required; `symbolic_rewriting` has no docstring.
```

